PR #13, opened 2015-06-10
fleet_units: add logrotate container

chr0n1x:


v1/fleet_units/logrotate.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Log Rotator for docker containers
+After=docker.service
+
+[Service]
```
cfortier2, line 6: can we add `User=core`

v1/fleet_units/logrotate.service
```diff
@@ -0,0 +1,17 @@
...
+Restart=always
+ExecStartPre=-/usr/bin/docker stop logrotate
+ExecStartPre=-/usr/bin/docker rm -f logrotate
+ExecStartPre=/usr/bin/docker pull michaloo/logrotate
+TimeoutStartSec=0
+ExecStartPre=-/usr/bin/docker kill logrotate
+ExecStartPre=-/usr/bin/docker rm logrotate
+ExecStart=/usr/bin/docker run --name logrotate -v /var/lib/docker:/var/lib/docker michaloo/logrotate
```
iderdik: Consider "forking" (or whatever the dockerhub equivalent is) so we own the container and control versioning?

v1/fleet_units/logrotate.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Log Rotator for docker containers
+After=docker.service
```
cfortier2, line 3: you also need a `Requires=docker.service`

v1/fleet_units/logrotate.service
cfortier2: by changing to User=core, this should launch with `sudo`

cfortier2:
we need to update the fleet_bootstrap.sh script and add this to the array of services to launch.

nice job!

cfortier2:
:+1:

v1/fleet_units/logrotate.service
```diff
@@ -0,0 +1,19 @@
...
+[Service]
+User=core
+Restart=always
+ExecStartPre=-/usr/bin/docker stop logrotate
+ExecStartPre=-/usr/bin/docker rm -f logrotate
+ExecStartPre=/usr/bin/docker pull michaloo/logrotate
+TimeoutStartSec=0
+ExecStartPre=-/usr/bin/docker kill logrotate
+ExecStartPre=-/usr/bin/docker rm logrotate
+ExecStart=sudo /usr/bin/docker run --name logrotate -v /var/lib/docker:/var/lib/docker michaloo/logrotate
```
ko-be: @iderdik 's comment got lost as this line changed, but I also agree w/ forking and making our own container so that we have full control of the contents of the container.  It's fine for now, but something we will need to do for most containers before Production.

chr0n1x: yes, working on that now - adding a few usability patches

v1/fleet_units/logrotate.service
```diff
@@ -0,0 +1,19 @@
...
+Description=Log Rotator for docker containers
+After=docker.service
+Requires=docker.service
+
+[Service]
+User=core
+Restart=always
+ExecStartPre=-/usr/bin/docker stop logrotate
+ExecStartPre=-/usr/bin/docker rm -f logrotate
+ExecStartPre=/usr/bin/docker pull michaloo/logrotate
```
cfortier2: you're still pulling michaloo.

might want to add something like:
`Environment=FD_IMAGE=behance/flight-director:latest`

then reference the variable in the pull/run commands

chr0n1x:
@cfortier2 @iderdik for the record - I confirmed that this works locally :D

PR #16, opened 2015-06-12
add a touch to the .dockercfg file

cfortier2:


ko-be:
if you want it owned as core:core you may also need a chown in there @cfortier2

PR #21, opened 2015-07-13
update cleanup to run every 4hours and only remove week old containers

cfortier2:


bryanlatten:
week old might be a....lot, no? especially since almost any host can run any container

cfortier2:
my thinking was that if theres an issue at 8pm on a friday, we'd want to be able to look at it on monday morning. so that gives us like 3 days. but trying to figure out 3 days in bash isn't fun, but 1 week is easy. 

but i can work it more so that we keep a couple of days.

bryanlatten:
ok, i'm only worried about the immediate disk IO as it removes a full week of containers. can we set it to a max that it would try to remove?

v1/fleet_units/docker-cleanup.service
```diff
@@ -9,15 +9,16 @@ Restart=on-failure
 RestartSec=20
 TimeoutStartSec=0
 ExecStart=/usr/bin/sh -c ' \
-  if [[ $(/usr/bin/docker ps -a | grep Exited | wc -l) -ne 0 ]]; \
+  if [[ $(/usr/bin/docker ps -a | grep Exited | awk \'{print $1 "\\t" $7 "\\t" $10}\' | grep week | wc -l) -ne 0 ]]; \
   then \
     echo "starting docker cleanup"; \
-    /usr/bin/docker ps -a | grep Exited | awk \'{print $1}\' | xargs /usr/bin/docker rm ; \
+    /usr/bin/docker ps -a | grep Exited | awk \'{print $1 "\\t" $7 "\\t" $10}\' | grep week | awk \'{print $1}\' | head -n25 | xargs /usr/bin/docker rm -v ; \
```
bryanlatten: can change the output to `starting docker cleanup, 25 max`? (using a new variable, to be defined above)

cfortier2: yes. added.

bryanlatten:
:+1:

bryanlatten:
lookin good!

PR #27, opened 2015-07-28
change units to Restart=always

cfortier2:


v1/fleet_units/docker-cleanup.service
```diff
@@ -7,7 +7,7 @@ Requires=docker.service
 Environment=MAX_DOCKER=25
 
 User=core
-Restart=on-failure
```
nBerg, line 10: Does this work with the timer thing?

cfortier2, line 10: oh, we probably don't need that there.

nBerg:
:+1:

PR #28, opened 2015-07-30
Convert docker-cleanup to a script for easier maintainability

cfortier2:


v1/fleet_units/docker-cleanup.service
```diff
@@ -10,17 +10,7 @@ User=core
...
-  then \
-    echo "Starting docker cleanup with a max of: $MAX_DOCKER"; \
-    /usr/bin/docker ps -a | grep Exited | awk \'{print $1 "\\t" $7 "\\t" $10}\' | grep week | awk \'{print $1}\' | head -n $MAX_DOCKER | xargs /usr/bin/docker rm -v ; \
-    /usr/bin/docker images -a | grep none | awk \'{print $3}\' | xargs /usr/bin/docker rmi ; \
-    echo "cleanup complete"; \
-  else \
-    echo "nothing to cleanup"; \
-  fi
-  sudo free -h'
+ExecStart=/usr/bin/bash -c '/usr/bin/wget -O /home/core/docker-cleanup.sh https://raw.githubusercontent.com/behance/mesos-systemd/fix/docker-cleanup/v1/fleet_units/docker-cleanup.sh && /usr/bin/bash /home/core/docker-cleanup.sh'
```
iderdik, line 13: Can you forsee any need to version this file? Once deployed, all clusters and hosts that use this will automatically use the latest version. That might be fine but versioning would give us a bit more control as to which clusters get which versions of the cleanup script.

PR #31, opened 2015-08-05
convert control services to templates 

cfortier2:
This allows us to treat each unit as a template for fleet so that we can control granularity of each service.


iderdik:
:+1:

PR #32, opened 2015-08-07
uses behance container

iderdik:
All this container does is:

```
FROM mbabineau/zookeeper-exhibitor:3.4.6_1.5.5
MAINTAINER Israel Derdik derdik@adobe.com

RUN ln -sf /dev/stdout /opt/zookeeper/zookeeper.out 
```

### DIFF
--- a/v1/fleet_units/zookeeper-exhibitor.service
+++ b/v1/fleet_units/zookeeper-exhibitor.service
@@ -4,7 +4,7 @@ After=docker.service
 Requires=docker.service
 
 [Service]
-Environment=EXHIBITOR_IMAGE=mbabineau/zookeeper-exhibitor:3.4.6_1.5.5
+Environment=EXHIBITOR_IMAGE=behance/exhibitor:latest
 
 User=core
 Restart=always


bryanlatten:
:+1:

chrishenry:
:+1:

chrishenry:
I feel like we should do this most of the OSS containers that we run. It always makes me nervous to rely on the OSS world to keep everything in exactly the same place we need them to be...

bryanlatten:
@chrishenry totally agreed. the only complication is public vs. private

v1/fleet_units/zookeeper-exhibitor.service
```diff
@@ -4,7 +4,7 @@ After=docker.service
 Requires=docker.service
 
 [Service]
-Environment=EXHIBITOR_IMAGE=mbabineau/zookeeper-exhibitor:3.4.6_1.5.5
+Environment=EXHIBITOR_IMAGE=behance/exhibitor:latest
```
cfortier2, line 7: this should probably be locked to a version

PR #33, opened 2015-08-10
Marathon uses zookeeper on localhost

iderdik:
Marathon only resolves host names once. So if an underlying ZK ELB node is replaced, Marathon continues to try to connect to the missing node. Using localhost will always provide a valid ZK endpoint. Testing shows no ill affects to cluster health if one node loses ZK.


nBerg:
thumb

PR #42, opened 2015-08-21
v2 & init script

chr0n1x:


v2/units/control-jenkins.service
```diff
@@ -0,0 +1,37 @@
...
+  -u root \
+  jenkins'
+ExecStop=/usr/bin/docker stop jenkins
+
+[Install]
+WantedBy=multi-user.target
+
+[X-Fleet]
+Global=false
+MachineMetadata=role=control
```
iderdik, line 37: Will it be a problem to have 1 on each control node?

chr0n1x, line 37: this only runs 1 container across all control nodes

iderdik:
:+1:

PR #54, opened 2015-08-28
mesosphere setup weee

chr0n1x:


v2/setup/mesosphere.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+source /etc/environment
+
+if [ -z $MESOS_SLAVE_IMAGE ]; then
```
nBerg, line 5: could we use this script to update the image? or do we use something else for that?

chr0n1x, line 5: that'll have to be something else I think

PR #55, opened 2015-08-28
Misc tweaks to mesos scripts

cfortier2:


v2/fleet/mesos-slave.service
```diff
@@ -8,14 +8,12 @@ Requires=docker.service
 EnvironmentFile=/etc/environment
 
 User=core
-KillMode=process
 Restart=always
 RestartSec=20
 TimeoutStartSec=0
+ExecStartPre=/usr/bin/docker pull ${MESOS_IMAGE}
```
chr0n1x: `MESOS_SLAVE_IMAGE`

v2/fleet/mesos-slave.service
```diff
@@ -33,7 +31,6 @@ ExecStart=/usr/bin/sh -c "sudo /usr/bin/docker run \
     --containerizers=docker,mesos \
     --executor_registration_timeout=10mins \
     --hostname=`curl -s http://169.254.169.254/latest/meta-data/public-hostname` \
-    --isolation=cgroups/cpu,cgroups/mem \
```
chr0n1x, line 36: what exactly does this do?

cfortier2, line 36: in our config, it doesn't seem to do anything. This was a recommendation from somewhere to get mesos-in-a-container to perform better. But It's not the default and seems to work fine without it, so  I'm removing it.

From the (sparse) documentation:

```
Isolation mechanisms to use, e.g., 'posix/cpu,posix/mem', or 'cgroups/cpu,cgroups/mem', or network/port_mapping (configure with flag: --with-network-isolator to enable), or 'external', or load an alternate isolator module using the --modules flag. Note that this flag is only relevant for the Mesos Containerizer. (default: posix/cpu,posix/mem)
```

chr0n1x, line 36: do we have...numbers to support that?

chr0n1x, line 36: I don't mind either way, just something to keep in mind since we're aiming for more stability

chr0n1x:
:+1:

v2/fleet/mesos-slave.service
cfortier2, line 36: not really. we don't have numbers either way. Switching back to the default _seems_ like the safer thing to do.

PR #60, opened 2015-09-02
Convert v2 services to use templates

cfortier2:


chr0n1x:
can these be run with just `fleetctl submit` & `start`?

cfortier2:
we need to append the number to it. So something like:

```
for i in {1..5};
do
  fleetctl start mesos-master@$i.service
done
```

chr0n1x:
for each running instance or instance that it needs to run on??

cfortier2:
the fleet command can be run from one host, but each service will need to be called n times for the number of control nodes that we want.

The full documentation is: https://coreos.com/fleet/docs/latest/launching-containers-fleet.html#run-a-high-availability-service

chr0n1x:
ah ok. the only issue with this setup is that now the `init` script needs to be updated to understand this. While _we_ may know that we have 5 nodes in the control tier, the script does not. We need a reliable way to determine that number

cfortier2:
we could probably pass in the control tier desired size?

chr0n1x:
that doesn't seem...clean.

cfortier2:
it's an option. or we could add a new parameter?

PR #63, opened 2015-09-03
v2: add dd-agent-mesos

chr0n1x:


v2/fleet/datadog-control.service
```diff
@@ -0,0 +1,27 @@
+# https://raw.githubusercontent.com/DataDog/docker-dd-agent/master/dd-agent.service
+# make sure the etcd /ddapikey is set!
+[Unit]
```
chrishenry, line 3: We also need `ddmesoshost` set in `etcd`. The value needs to be the control ELB.

v2/fleet/datadog-control.service
```diff
@@ -0,0 +1,27 @@
...
+ExecStartPre=-/usr/bin/docker stop dd-agent-mesos
+ExecStartPre=-/usr/bin/docker rm -f dd-agent-mesos
+TimeoutStartSec=0
+ExecStartPre=-/usr/bin/docker kill dd-agent-mesos
+ExecStartPre=-/usr/bin/docker rm dd-agent-mesos
+ExecStart=/usr/bin/bash -c \
+"if [[ -f /etc/profile.d/etcdctl.sh ]]; then source /etc/profile.d/etcdctl.sh;fi && sudo /usr/bin/docker run --name dd-agent-mesos -h `hostname` \
+-v /var/run/docker.sock:/var/run/docker.sock \
+-v /proc/mounts:/host/proc/mounts:ro \
+-v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro \
```
chrishenry: We actually don't need or want any of these mounts. These mounts are for datadog to grab information from docker etc.

In this case, this container is _only_ responsible for pulling stats on mesos, marathon and zookeeper.

v2/fleet/datadog-control.service
```diff
@@ -0,0 +1,27 @@
...
+-e API_KEY=`etcdctl get /ddapikey` \
```
chrishenry, line 20: We also need a  `-e API_KEY=`etcdctl get /ddmesoshost`

v2/fleet/datadog-control.service
```diff
@@ -0,0 +1,27 @@
...
+behance/docker-dd-agent-mesos"
+ExecStop=/usr/bin/docker stop dd-agent-mesos
+
+[X-Fleet]
+Global=false
+MachineMetadata=role=control
```
chrishenry, line 27: It doesn't actually _need_ to be on the control hosts, it just needs to be somewhere where it can get to ddmesoshost / control elb. However, it's certainly OK to run it there.

Might be helpful if we ever decide / need to move it somewhere else.

chr0n1x, line 27: Just going to leave it here for now. That way we can immediately get stats as soon as the control tier spins up, instead of having to wait for any other tier. We are getting stats of things that only run on the control tier, after all.

v2/fleet/datadog-control.service
```diff
@@ -0,0 +1,26 @@
...
+ExecStartPre=/usr/bin/docker pull behance/docker-dd-agent-mesos:latest
+ExecStartPre=-/usr/bin/docker stop dd-agent-mesos
+ExecStartPre=-/usr/bin/docker rm -f dd-agent-mesos
+TimeoutStartSec=0
+ExecStartPre=-/usr/bin/docker kill dd-agent-mesos
+ExecStartPre=-/usr/bin/docker rm dd-agent-mesos
+ExecStart=/usr/bin/bash -c \
+"if [[ -f /etc/profile.d/etcdctl.sh ]]; then source /etc/profile.d/etcdctl.sh;fi && sudo /usr/bin/docker run --name dd-agent-mesos -h `hostname` \
+-e API_KEY=`etcdctl get /ddapikey` \
+-e MESOS_HOST \
```
chr0n1x: @chrishenry does this just need to hostname? Or do we also need the port that mesos is running on too? I can just patch the CF templates to write `MESOS_HOST=...` to `/etc/environment` and use it here.

v2/fleet/datadog-control.service
```diff
@@ -0,0 +1,26 @@
...
+[Service]
+EnvironmentFile=/etc/environment
+User=core
+Restart=always
+ExecStartPre=/usr/bin/docker pull behance/docker-dd-agent-mesos:latest
+ExecStartPre=-/usr/bin/docker stop dd-agent-mesos
+ExecStartPre=-/usr/bin/docker rm -f dd-agent-mesos
+TimeoutStartSec=0
+ExecStartPre=-/usr/bin/docker kill dd-agent-mesos
+ExecStartPre=-/usr/bin/docker rm dd-agent-mesos
```
cfortier2: this line is a duplicate of line 13

v2/fleet/datadog-control.service
chrishenry: Nope, because I'm monitoring mesos, zookeeper and marathon, I append the ports in the [entrypoint script](https://github.com/behance/docker-dd-agent-mesos/blob/master/entrypoint.sh#L11). Not as flexible should we decide to change the ports, but it's easy enough to redeploy the container at this point.

v2/fleet/datadog-control.service
chrishenry, line 27: Yerp, makes sense.

v2/fleet/datadog-control.service
chr0n1x: ah ok, I'll have to update the CF templates either way, will get to it soon

v2/fleet/datadog-control.service
```diff
@@ -0,0 +1,24 @@
...
+User=core
+Restart=always
+TimeoutStartSec=0
+ExecStartPre=/usr/bin/docker pull behance/docker-dd-agent-mesos:latest
+ExecStartPre=-/usr/bin/docker kill dd-agent-mesos
+ExecStartPre=-/usr/bin/docker rm -f dd-agent-mesos
+ExecStart=/usr/bin/bash -c \
+"if [[ -f /etc/profile.d/etcdctl.sh ]]; then source /etc/profile.d/etcdctl.sh;fi && sudo /usr/bin/docker run --name dd-agent-mesos -h `hostname` \
+-e API_KEY=`etcdctl get /ddapikey` \
+-e MESOS_HOST \
```
chrishenry: So the control ELB will also need to be set in etcd. Something along the lines of;

```
-e MESOS_HOST=`etcdctl get /ddmesoshost`
```

``` bash
chris@ip-10-43-1-13 stage-portfolio-us-west-2a-control ~ $ etcdctl get /ddmesoshost
internal-portfolio-control-elb-name-other-stuff.us-south-8.elb.amazonaws.com
```

chrishenry:
:+1:

PR #76, opened 2015-09-18
have capcom run in privileged mode

nBerg:


chrishenry:
GIVE ALL TEH PRIVILEGE!!!!1!!

Actually, what's the reason we need to run in `privileged` mode?

nBerg:
We think we may have found a way to prevent the haproxy restart throwing 503s, but it requires messing with iptable rules

chrishenry:
ohh, I saw the other PR...so we're manipulating iptables on the CoreOS host via the container?

nBerg:
yupp

iderdik:
:+1:

mannytoledo:
notpass

PR #79, opened 2015-09-21
Envrionment setup script

nBerg:
@chr0n1x


chrishenry:
What is this for?

nBerg:
Changing values in this file requires an A/B deploy. Pulling values from etcd vs this file makes updating the values on the fly super easy.

chr0n1x:
:+1:

PR #80, opened 2015-09-21
Proxy setup systemd unit

iderdik:
Runs https://github.com/behance/mesos-proxy-setup once on start up.


chr0n1x:
@iderdik I would actually just turn this into a setup script with:

```
if [ "$NODE_ROLE" != "proxy" ]; then
    exit 0
fi
```

at the top

chr0n1x:
:+1:

PR #83, opened 2015-09-23
etcd for ELB configurations

chr0n1x:


v2/fleet-local/control/mesos-master@.service
```diff
@@ -26,7 +26,7 @@ ExecStart=/usr/bin/sh -c "sudo /usr/bin/docker run \
   --log_dir=/var/log/mesos \
   --quorum=3 \
   --work_dir=/var/lib/mesos/master \
-  --zk=zk://${ZOOKEEPER_ENDPOINT}/mesos"
+  --zk=zk://`etcdctl get /environment/ZOOKEEPER_ENDPOINT`/mesos"
```
ko-be, line 29: You might want to use the $() syntax instead of backticks to keep things consistent @chr0n1x unless that syntax doesn't work for running etcdctl.

ko-be:
Otherwise :+1:

v2/fleet-local/control/mesos-master@.service
chr0n1x, line 29: for the most part, I believe that shell-outs in bash `-c` commands are backticks. but still, noted. It'll be one of the things that we standardize when we clean this a bit more.

PR #84, opened 2015-09-24
dd-agent unit for proxy tier

chr0n1x:


v2/fleet/datadog.service
```diff
@@ -22,3 +22,4 @@ ExecStop=/usr/bin/docker stop dd-agent
 
 [X-Fleet]
 Global=true
+MachineMetadata=role=worker
```
chrishenry: Wouldn't we want to run this on the proxy tier?

It looks like both the [haproxy](https://github.com/behance/docker-dd-agent-proxy/blob/master/conf.d/haproxy.yaml#L4) and [nginx](https://github.com/behance/docker-dd-agent-proxy/blob/master/conf.d/nginx.yaml#L13) data integrations are set up to look at `localhost`.

PR #86, opened 2015-09-25
fix: Clean up images regardless of Exited containers

chrishenry:


v2/util/docker-cleanup.sh
```diff
@@ -6,9 +6,12 @@ if [[ $(/usr/bin/docker ps -a | grep Exited | awk '{print $1 "\\t" $7 "\\t" $10}
...
   /usr/bin/docker ps -a | grep Exited | awk '{print $1 "\\t" $7 "\\t" $10}' | grep week | awk '{print $1}' | head -n $MAX_DOCKER | xargs /usr/bin/docker rm -v
-  /usr/bin/docker images -a | grep none | awk '{print $3}' | xargs /usr/bin/docker rmi
-  echo "cleanup complete"
+  echo "container cleanup complete"
 else
-  echo "nothing to cleanup"
+  echo "no containers to cleanup"
 fi
+
+/usr/bin/docker images -a | grep none | awk '{print $3}' | xargs /usr/bin/docker rmi
```
ko-be: we should get rid of `| grep none` @chrishenry

v2/util/docker-cleanup.sh
```diff
@@ -6,9 +6,12 @@ if [[ $(/usr/bin/docker ps -a | grep Exited | awk '{print $1 "\\t" $7 "\\t" $10}
...
   /usr/bin/docker ps -a | grep Exited | awk '{print $1 "\\t" $7 "\\t" $10}' | grep week | awk '{print $1}' | head -n $MAX_DOCKER | xargs /usr/bin/docker rm -v
-  /usr/bin/docker images -a | grep none | awk '{print $3}' | xargs /usr/bin/docker rmi
-  echo "cleanup complete"
+  echo "container cleanup complete"
 else
-  echo "nothing to cleanup"
+  echo "no containers to cleanup"
 fi
+
+/usr/bin/docker images -a  | awk '{print $3}' | xargs /usr/bin/docker rmi
```
chr0n1x, line 14: so just remove everything that's not being used?
what if docker updates to allow `-f` by default for the `rmi` command? (sorry I'm being paranoid about this)

cfortier2, line 14: this PR is fine, but probably won't help much. you can't remove an image if there's still a container that relies on it, which is why the `docker rmi` command is wrapped in the `if` block.

see:
image

cfortier2, line 14: though you could also simplify it to be:

`docker rmi $(docker images -aq)`

to simply attempt deleting all images.

chrishenry:
@chr0n1x If docker updates to adding the `-f` by default, it would go against every sort of sane behaviour I've ever observed in any program that implements a force flag. If that happens, it would be my recommendation to abandon Docker as a platform safe to run in production.

What do you think the chances of them changing that behaviour are?

v2/util/docker-cleanup.sh
ko-be, line 14: So @cfortier2 this PR is a result of one of the nodes running out of disk space.  Upon investigation, we found nearly 40GB of docker images that were no longer used, and running the above command fixed it.

The original command also had a `| grep none` in it, which caused it to miss quite a few images.

ko-be:
@chr0n1x @chrishenry agreed, that is not something docker should be changing defaults for, certainly not between minor versions.

chr0n1x:
@chrishenry @ko-be fair point, I can just see a way to do it based on the current running containers that could avoid edge cases like that.

`docker ps|grep -v IMAGE|tr -s [:space:]|cut -d' ' -f2` & `grep -v` the generated list.

But whatever works for the time being I guess.

anshupande:
https://forums.docker.com/t/command-to-remove-all-unused-images/20

anshupande:
$ docker images -f dangling=true -q | xargs docker rmi

If you are using Dockerfiles to build your Images (recommended) then you can always rebuild an image should anything get removed accidentally.

anshupande:
Also to clean up non-running containers

$ docker ps -a --no-trunc | grep 'Exit' | awk '{print $1}' | xargs docker rm

ko-be:
We should consider using https://github.com/spotify/docker-gc or https://github.com/yelp/docker-custodian rather than rolling our own.

anshupande:
Docker bench also has good reference, look out for 6.6 Avoid image sprawl (page 111 ,112 and 113) : https://benchmarks.cisecurity.org/tools2/docker/CIS_Docker_1.6_Benchmark_v1.0.0.pdf

chr0n1x:
@ko-be definitely a fan of [docker-custodian](https://github.com/Yelp/docker-custodian). That's some good python code, AND they have a docker solution.

cfortier2:
Here's an interesting comment in the spotify solution to be aware of: https://github.com/spotify/docker-gc/blob/master/docker-gc#L27

chrishenry:
Good suggestions here. I'm going to merge this.

I'll leave it up to anyone who suggested changes here to execute on them.

PR #93, opened 2015-10-02
Docker-Cleanup automation tweaks

chr0n1x:
- only run on worker nodes
- remove exited containers & volumes, per usual
- do a basic run-through of `docker rmi` (not `--force` so that running containers stay running)
- audit `/var/lib/docker/$FSDRIVER`, cross-reference with daemon and delete

### DIFF
--- a/v2/fleet/docker-cleanup.service
+++ b/v2/fleet/docker-cleanup.service
@@ -4,8 +4,6 @@ After=docker.service
 Requires=docker.service
 
 [Service]
-Environment=MAX_DOCKER=25
-
 User=core
 Restart=on-failure
 RestartSec=20
--- a/v2/fleet/docker-cleanup.service
+++ b/v2/fleet/docker-cleanup.service
@@ -17,4 +17,3 @@ https://raw.githubusercontent.com/behance/mesos-systemd/master/v2/util/docker-cl
 [X-Fleet]
 Global=true
 MachineMetadata=role=worker
-MachineMetadata=role=proxy
--- a/v2/fleet/docker-cleanup.service
+++ b/v2/fleet/docker-cleanup.service
@@ -10,9 +10,7 @@ User=core
 Restart=on-failure
 RestartSec=20
 TimeoutStartSec=0
-ExecStart=/usr/bin/bash -c '/usr/bin/wget -O /home/core/docker-cleanup.sh \
-https://raw.githubusercontent.com/behance/mesos-systemd/master/v2/util/docker-cleanup.sh \
-&& /usr/bin/bash /home/core/docker-cleanup.sh'
+ExecStart=/usr/bin/sudo bash /home/core/mesos-systemd/v2/util/docker-cleanup.sh
 
 [X-Fleet]
 Global=true
--- a/v2/util/docker-cleanup.sh
+++ b/v2/util/docker-cleanup.sh
@@ -1,17 +1,27 @@
 #!/bin/bash
 
-# this exists because trying to escape bash commands in a sytemd unit is super sadtimes
+# NOTE: this script will need to be sudo'ed for access to /var/lib/docker
 
-if [[ $(/usr/bin/docker ps -a | grep Exited | awk '{print $1 "\\t" $7 "\\t" $10}' | grep week | wc -l) -ne 0 ]]
-then
-  echo "Starting docker cleanup with a max of: $MAX_DOCKER"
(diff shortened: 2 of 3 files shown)


v2/fleet/docker-cleanup.service
```diff
@@ -10,11 +10,8 @@ User=core
 Restart=on-failure
 RestartSec=20
 TimeoutStartSec=0
-ExecStart=/usr/bin/bash -c '/usr/bin/wget -O /home/core/docker-cleanup.sh \
-https://raw.githubusercontent.com/behance/mesos-systemd/master/v2/util/docker-cleanup.sh \
-&& /usr/bin/bash /home/core/docker-cleanup.sh'
+ExecStart=/usr/bin/sudo bash /home/`whoami`/mesos-systemd/v2/util/docker-cleanup.sh
```
ko-be: How can the `whoami` call possibly work here?  I'd say it's better to hardcode `/home/core` @chr0n1x

ko-be: @chr0n1x pointed out that User=core is defined above this line.  Does `$HOME/mesos-systemd/v2/util/docker-cleanup.sh` work?

chr0n1x: nope, unless we run in through `/usr/bin/bash -c`

v2/util/docker-cleanup.sh
```diff
@@ -1,17 +1,27 @@
...
-else
-  echo "no containers to cleanup"
-fi
+echo "Removing dead containers & volumes"
+docker rm -v $(docker ps -a|grep Exited|cut -d" " -f1) 2> /dev/null
 
-/usr/bin/docker images -a  | awk '{print $3}' | xargs /usr/bin/docker rmi
-echo "image cleanup complete"
+echo "Removing images"
+docker rmi $(docker images -aq) 2> /dev/null
```
anshupande, line 10: are we planning to remove all currently instantiated images or only images that are not instantiated

chr0n1x, line 10: @anshupande it's not a `--force` so running images will not get deleted.

v2/util/docker-cleanup.sh
```diff
@@ -1,17 +1,27 @@
 #!/bin/bash
 
-# this exists because trying to escape bash commands in a sytemd unit is super sadtimes
+# NOTE: this script will need to be sudo'ed for access to /var/lib/docker
```
cfortier2, line 3: if we need sudo access, we could simply run this as `root`?

chr0n1x, line 3: I thought we were standardizing as `core` and only elevating, explicitly, when needed.

cfortier2, line 3: yes. I think standardizing as `core` is probably the right way to go. But if it starts to get too messy, we can run it as root if we need to.

cfortier2:
Do we no longer want to keep stopped containers for awhile in case we need to debug them?

chr0n1x:
@cfortier2 I guess that the answer to that question is in https://github.com/behance/mesos-systemd/pull/86 :/

cfortier2:
no, the check to remove containers that exited over a week ago is being removed in this PR. It's currently on line #8 of master: https://github.com/behance/mesos-systemd/blob/master/v2/util/docker-cleanup.sh#L8

ko-be:
@cfortier2 I think a week is a bit too long to keep old containers, as we filled around 40gb in 2 days.

Also, doesn't this line just remove all unused images, anyways?  https://github.com/behance/mesos-systemd/blob/master/v2/util/docker-cleanup.sh#L14

cfortier2:
@ko-be we decided to keep them at least 4 days. Reasoning is that if a container fails at 8pm on Friday, but Monday is a holiday, we want to keep it around for post-mortem on Tuesday. But checking for `week` is way simpler than bash magic for `n` number of days. 

No, that line shouldn't remove an image if it's still being used by an Exited container.

cfortier2:
@ko-be also, disk size should be 250gb now. Storage is cheap, so it should be able to up the disks on these if needed. Our dev laptops have more disk space than these production servers.

ko-be:
@cfortier2 That sounds reasonable to me.  If we're going to keep all container images around for a week, I'd say we bump up the disk space even more to prevent them from running out.

chr0n1x:
merging this. the amount of crap from clobbered fs layers right now is making everything horrible

PR #95, opened 2015-10-07
zk-health.service to block locksmithd, hardening unit dependencies

chr0n1x:
- block `locksmithd` startup based on `/v1/exhibitor/cluster/status`
- have marathon & mesos services wait for `zk-health.service` to complete before starting

Note that these changes only affects unit **startup**. On purpose, I omitted using `PartOf=` & `BindsTo=` relations. This way, if things like `zk-health` disappears all of a sudden the `stop` or `restart` actions do not cascade down to child units. However, we might want to evaluate those at some point in time.
### On initial boot/startup

Before zk-exhibitor has started & have all `etcdctl get /zookeeper/ensemble_size` nodes serving:
screen shot 2015-10-09 at 2 46 53 pm

After all ZK members start serving:
screen shot 2015-10-09 at 2 51 03 pm
### on reboot

before everything starts running
screen shot 2015-10-09 at 2 55 54 pm

after ZK joins the ensemble. Note that locksmithd starts up (timestamps indicate this)
screen shot 2015-10-09 at 2 57 14 pm

### DIFF
--- a/v2/fleet-local/control/flight-director@.service
+++ b/v2/fleet-local/control/flight-director@.service
@@ -2,9 +2,9 @@
 Description=Flight Director @ %i
 # it's implied that chronos & marathon require mesos
 # look here: https://github.com/behance/mesos-systemd/tree/master/v2/fleet
-After=docker.service chronos.service marathon.service
+After=docker.service chronos@%i.service marathon@%i.service
 Requires=docker.service
-Wants=chronos.service marathon.service
+Wants=chronos@%i.service marathon@%i.service
 
 [Service]
 Environment="IMAGE=etcdctl get /images/fd"
--- a/v2/fleet-local/control/flight-director@.service
+++ b/v2/fleet-local/control/flight-director@.service
@@ -10,8 +10,10 @@ Wants=chronos@%i.service marathon@%i.service
 Environment="IMAGE=etcdctl get /images/fd"
 User=core
 Restart=on-failure
-RestartSec=20
+RestartSec=8
 TimeoutStartSec=0
+
+ExecStartPre=/usr/bin/systemctl is-active marathon@%i.service
 ExecStartPre=/usr/bin/sh -c "docker pull $($IMAGE)"
 ExecStartPre=-/usr/bin/docker kill flight-director
 ExecStartPre=-/usr/bin/docker rm flight-director
--- a/v2/fleet-local/control/zk-health@.service
+++ b/v2/fleet-local/control/zk-health@.service
@@ -1,0 +1,33 @@
+[Unit]
+Description=Exhibitor/Zookeeper health checker/block @ %i
+
+# forces things like the update-engine (and related services such as managed
+# CoreOS) to wait for this unit to complete
+Before=locksmithd.service marathon@%i.service mesos-master@%i.service
+After=zk-exhibitor@%i.service
+
+[Service]
+Type=simple
+User=core
+Restart=on-failure
+RemainAfterExit=true
+
+# array of objects, each representing a ZK node
+Environment="EX_QUERY=curl -sS localhost:8181/exhibitor/v1/cluster/status"
+# expected size of ensemble after it's done
+Environment="CTL_SIZE=etcdctl get /environment/CONTROL_CLUSTER_SIZE"
+
+# wait for zk-exhibitor to start - can't use %i specifier here
+ExecStartPre=/usr/bin/bash -c 'while [ "$(systemctl is-active zk-exhibitor@*)" != "active" ]; do sleep 2; done'
+# code==3 means that the node is up and serving
+ExecStartPre=/usr/bin/bash -c 'while [ $($EX_QUERY|jq "[select(.[].code == 3)]|length") != $($CTL_SIZE) ]; do sleep 4; done'
+
+ExecStart=/usr/bin/sleep infinity
+
+[Install]
+WantedBy=multi-user.target
+
+[X-Fleet]
+Global=true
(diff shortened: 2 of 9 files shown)


v2/fleet/zk-health.service
```diff
@@ -0,0 +1,30 @@
...
+Type=oneshot
+User=core
+
+# on-success - allows other units to reference state of this unit
+RemainAfterExit=true
+
+# array of objects, each representing a ZK node
+Environment="EX_QUERY=curl -sS localhost:8181/exhibitor/v1/cluster/status"
+# code == 3 ==> node is serving
+Environment="JQ_CMD=docker run -i --rm dtolnay/jq '[select(.[].code == 3)]|length'"
```
chrishenry: isn't `jq` installed natively on CoreOS? I think it is.

```
chris@ip-10-131-101-71 prod-pro2-us-east-1c-control ~ $ which jq
/usr/bin/jq
```

Also, sweet! jq!

v2/fleet/zk-health.service
```diff
@@ -0,0 +1,30 @@
...
+# expected size of ensemble after it's done
+Environment="CTL_SIZE=etcdctl get /environment/CONTROL_CLUSTER_SIZE"
+
+ExecStartPre=/usr/bin/bash -c "while [ $($EX_QUERY|$JQ_CMD) != $($CTL_SIZE) ]; do sleep 4; done"
+ExecStart=/usr/bin/bash -c "ZK instances ok! $($EX_QUERY)|docker run -i --rm dtolnay/jq '.'"
```
chrishenry: This is a really cool pattern.

v2/fleet/zk-health.service
chr0n1x: WHOA IT IS :+1: :boom:

mannytoledo:
I love this check. This is really elegant, :heart_eyes:

v2/fleet-local/control/zk-health@.service
```diff
@@ -0,0 +1,35 @@
...
+
+# array of objects, each representing a ZK node
+Environment="EX_QUERY=curl -sS localhost:8181/exhibitor/v1/cluster/status"
+# expected size of ensemble after it's done
+Environment="CTL_SIZE=etcdctl get /environment/CONTROL_CLUSTER_SIZE"
+
+# wait for zk-exhibitor to start - can't use %i specifier here
+ExecStartPre=/usr/bin/bash -c 'while [ "$(systemctl is-active zk-exhibitor@*)" != "active" ]; do sleep 2; done'
+# give the instance ~16s to start up & serve since the container starting doesn't mean much for curl
+ExecStartPre=/usr/bin/sleep 16
```
chrishenry: Why 16 seconds? feels  kinda random

chr0n1x: power of 2. and 8 seconds was not enough

chrishenry: Okie dokie, makes sense

chr0n1x: This unit will restart if it does fail either way, so technically it's not needed. But I wanted to avoid having this unit restart over and over, spamming the logs post-reboot

chr0n1x:
I just realized that this needs to go into the userdata instead or something of the like, or else it does not persist on the machine.

v2/fleet-local/control/flight-director@.service
```diff
@@ -2,16 +2,18 @@
 Description=Flight Director @ %i
 # it's implied that chronos & marathon require mesos
 # look here: https://github.com/behance/mesos-systemd/tree/master/v2/fleet
-After=docker.service chronos.service marathon.service
+After=docker.service chronos@%i.service marathon@%i.service
```
nBerg, line 5: what if chronos never starts? will this block?

v2/fleet-local/control/flight-director@.service
```diff
@@ -2,16 +2,18 @@
...
 
 [Service]
 Environment="IMAGE=etcdctl get /images/fd"
 User=core
 Restart=on-failure
-RestartSec=20
+RestartSec=8
 TimeoutStartSec=0
+
+ExecStartPre=/usr/bin/systemctl is-active marathon@%i.service
```
nBerg, line 16: this is a thing??? omg yay

v2/fleet-local/control/flight-director@.service
chr0n1x, line 5: no, it's more of a `if there is a chronos@%i.service on this machine, schedule this unit to start after it`. What _really_ does the blocking is on line 16.

chr0n1x, line 5: or at least, I'm pretty sure that this is the case, as FD was able to start when I was testing

nBerg:
:+1:

mannytoledo:
:+1:

v2/fleet-local/control/flight-director@.service
chrishenry, line 5: Yea, this one is important, as we stopped enabling chronos on stage and prod bc it caused problems.

chr0n1x, line 5: Yeah. I was testing against a fresh control tier though and chronos was never started yet FD did. So I'm pretty sure that it's functioning correctly.

I did consciously leave this here though, as FD should not be started until after ALL mesos frameworks start up. This way, in clusters where we _do_ use chronos, systemd at least attempts to take its presence into account when starting the FD unit

chrishenry:
:+1:

PR #105, opened 2015-10-22
operation stability tweaks

chr0n1x:
- `ExecStop` commands that perform `docker stop` are allowed to fail (not all apps can be stopped cleanly)
- mesos-slave@ has an explicit check for the existence of `/home/core/.dockercfg` as a file
- template unit files remove themselves using an `ExecStop` option & the specifier


mannytoledo:
:+1:

nBerg:
:+1:

iderdik:
:+1:

PR #106, opened 2015-10-26
Separating Capcom from reverse proxy server

iderdik:
- Added new proxy@i.service 
- Setup nginx image

### DIFF
--- a/v2/fleet-local/proxy/capcom@.service
+++ b/v2/fleet-local/proxy/capcom@.service
@@ -10,6 +10,7 @@ RestartSec=20
 TimeoutStartSec=0
 Environment="IMAGE=etcdctl get /images/capcom"
 
+ExecStartPre=/usr/bin/systemctl is-active proxy@.service
 ExecStartPre=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && docker pull $($IMAGE)"
 ExecStartPre=-/usr/bin/docker kill capcom
 ExecStartPre=-/usr/bin/docker rm capcom
--- a/v2/fleet-local/proxy/capcom@.service
+++ b/v2/fleet-local/proxy/capcom@.service
@@ -21,6 +21,8 @@ ExecStart=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && \
     --name capcom \
     --net='host' \
     --privileged \
+    -v /etc/nginx:/etc/nginx \
+    -v /var/run/docker.sock:/var/run/docker.sock \
     -e LOG_APP_NAME=capcom \
     -e CP_APPLICATIONS=`etcdctl get /CP/CP_APPLICATIONS` \
     -e CP_HOST=`etcdctl get /CP/CP_HOST` \
--- a/v2/fleet-local/proxy/capcom@.service
+++ b/v2/fleet-local/proxy/capcom@.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Capcom @ %i
-After=docker.service proxy@%i.service
+After=docker.service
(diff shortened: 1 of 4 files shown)


v2/fleet-local/proxy/proxy@.service
```diff
@@ -0,0 +1,33 @@
...
+    $($IMAGE) nginx -g \"daemon off;\""  #HOW DO WE MAKE THIS CONDITIONAL? etcd?
+
+ExecStop=-/usr/bin/docker stop proxy
+ExecStop=/usr/bin/fleetctl destroy proxy@%i
+
+[Install]
+WantedBy=multi-user.target
+
+[X-Fleet]
+Global=false
```
chr0n1x: `true`

v2/fleet-local/proxy/proxy@.service
```diff
@@ -0,0 +1,33 @@
+[Unit]
+Description=Proxy { Nginx/HAProxy } @ %i
+After=docker.service
```
chr0n1x: I would add an explicit `Before=capcom@%i.service`

v2/fleet-local/proxy/proxy@.service
```diff
@@ -0,0 +1,33 @@
...
+ExecStartPre=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && docker pull $($IMAGE)"
+ExecStartPre=-/usr/bin/docker kill proxy
+ExecStartPre=-/usr/bin/docker rm proxy
+
+# NOTE: it's critical to source the etcdctl.sh file so that etcd connects to the correct cluster.
+ExecStart=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && \
+  docker run \
+    --name proxy \
+    --net='host' \
+    $($IMAGE) nginx -g \"daemon off;\""  #HOW DO WE MAKE THIS CONDITIONAL? etcd?
```
chr0n1x: yes

v2/fleet-local/proxy/proxy@.service
```diff
@@ -0,0 +1,33 @@
+[Unit]
+Description=Proxy { Nginx/HAProxy } @ %i
+After=docker.service
+Requires=docker.service
+
+[Service]
+User=core
+Restart=always
+RestartSec=20
```
chr0n1x: 20s? or less?

v2/fleet-local/proxy/proxy@.service
chr0n1x: nitpick: would using single quotes make this a bit more readable?

v2/fleet-local/proxy/capcom@.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
-Description=Capcom/{HAProxy NGinx} @ %i
-After=docker.service
+Description=Capcom @ %i
+After=docker.service proxy@%i.service
 Requires=docker.service
 
 [Service]
```
chr0n1x, line 7: In this `[Service]` block I would add an `ExecStartPre=/usr/bin/systemctl is-active proxy@*`

That will explicitly block this unit from starting up if capcom is not running

v2/fleet-local/proxy/proxy@.service
iderdik: What key namespace should we use in etcd? Does /CP/ make sense here even though its a sidekick service?

chr0n1x: I guess because this service is explicitly dependant on capcom... `/capcom/proxy`? We should definitely talk about a standard at some point

PR #110, opened 2015-11-03
v2/fleet: force services to wait for bootstrap

chr0n1x:
testing...


chr0n1x:
tested. seems to be resolving the FS layer clobbering issues on the slaves

PR #111, opened 2015-11-03
fleet/sumo*: --hostname flags to use public ipv4

chr0n1x:


ko-be:
Awesome @chr0n1x !!! :+1:

PR #114, opened 2015-11-17
memory limit for proxy

iderdik:
Prevents resource starvation (memory) by proxy


v2/fleet-local/proxy/proxy@.service
```diff
@@ -22,6 +22,7 @@ ExecStart=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && \
   echo $($CMD) | xargs docker run \
     --name proxy \
     --net='host' \
+    -m 12G \
```
chrishenry, line 25: Without the `-m` flag, how much memory does a container get by default?

iderdik, line 25: As much as it wants. As per CoreOS, this might be casuing starvation of sshd.

iderdik:
verbal thumb from @nBerg

PR #117, opened 2015-11-25
etcd2 unit files and scripts

nBerg:


init
```diff
@@ -1,21 +1,40 @@
...
+fi
+
+if [ "$NODE_ROLE" = "control" ]; then
+  echo "-------Control node found, setting up etcd peers-------"
+  sudo bash ${SCRIPTDIR}/v2/util/etcd-peers-dropin.sh
+  sudo systemctl enable "${SCRIPTDIR}/v2/util-units/etcd-peers.service"
+  sudo systemctl start etcd-peers
+
+  echo "-------Waiting for etcd2 to start-------"
+  until etcdctl cluster-health
```
chr0n1x: coooool

init
```diff
@@ -1,21 +1,40 @@
...
+  do
+    sleep 10
```
chr0n1x: why 10? why not...2, 4, or 8...? or another power of 2 :trollface:

nBerg: Cuz my brain is a base 10 system. I can change to whatever you want

init
```diff
@@ -1,21 +1,40 @@
...
+bash ${SCRIPTDIR}/v2/util/setup-credentials.sh
+
+if [ -f /etc/environment ]; then
+    source /etc/environment
+fi
+
+if [ "$NODE_ROLE" = "control" ]; then
+  echo "-------Control node found, setting up etcd peers-------"
+  sudo bash ${SCRIPTDIR}/v2/util/etcd-peers-dropin.sh
+  sudo systemctl enable "${SCRIPTDIR}/v2/util-units/etcd-peers.service"
```
chr0n1x: I think that it might be better to move or `cp` this into a dropin dir in `/etc/systemd`. The repo in `/home/core` _can_ be deleted at any point and `systemctl` symlinks the files, I think?

chr0n1x: also, I would go one step further and move all of this logic into `v2/util/etcd-peers-dropin.sh`. That's more for aesthetic reasons, but it makes this init file a bit more readable.

PR #120, opened 2015-11-30
Marathon auth

anshupande:


nBerg:
:+1: assuming the .marathon files exist in all the tiers

PR #131, opened 2015-12-14
v2/setup: prevent last line being ignored

fraczles:
`read` returns non-zero error code if it encounters EOF. This will check if `$line` is non-empty even if EOF is encountered.


nBerg:
lol :+1:

chr0n1x:
hahahaha :+1:

PR #142, opened 2016-02-12
Force Update CoreOS on boot.

chr0n1x:
- `update-os` fleet unit to force system OS update & reboot
- when there are no updates or if `update_engine_client` fails to `-update`, the unit just stays active instead of rebooting the system.
- mesos & marathon units require the `update-os` unit to be up & running.
- on cluster creation, this may cause failures as a system reboot will stall out mesos & marathon form starting up, and consequently, `zk-health`; ultimately, this will FORCE us to correctly update AMIs on cluster creation
- because this is a fleet unit, new nodes that are spinning up will automatically have this run. While it's running, that should prevent mesos & marathon from running too.

**NOTE**: this does NOT account for connectivity issues between the node & the update channel/provider

### DIFF
--- a/v2/fleet-local/control/marathon@.service
+++ b/v2/fleet-local/control/marathon@.service
@@ -14,6 +14,7 @@ Restart=always
 RestartSec=8
 TimeoutStartSec=0
 
+ExecStartPre=/usr/bin/systemctl is-active update-os.service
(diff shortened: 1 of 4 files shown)


nBerg:
:+1:

nBerg:
:+1:

PR #145, opened 2016-02-20
Add fleet service to login to AWS ECR. 

cdmurph32:
Add fleet service to login to AWS ECR. 
**This change also starts docker-cleanup.timer.**  I'm not sure if that was
intentionally deactivated in the switch from v1 to v2.

Please let me know if it would be better for the AWS ECR service and
timer file to live in fleet-local/worker.


init
```diff
@@ -42,7 +42,7 @@ do
     sudo fleetctl start "${unit%.service}${COREOS_PRIVATE_IPV4}"
 done
 # general-purpose units
-for unit in $(ls ${SCRIPTDIR}/v2/fleet|grep -e '.service$')
+for unit in $(ls ${SCRIPTDIR}/v2/fleet|grep -e '.service$\|.timer$')
```
chr0n1x, line 45: oh dang, good catch!

v2/fleet/aws-ecr-login.service
```diff
@@ -0,0 +1,22 @@
...
+[Service]
+Type=oneshot
+User=core
+Environment="AWS_ACCOUNT_ID=etcdctl get /ECR/registry-account"
+Environment="REGION=etcdctl get /ECR/region"
+ExecStartPre=/usr/bin/docker pull behance/utility:latest
+ExecStartPre=-/usr/bin/docker kill ecr-login
+ExecStartPre=-/usr/bin/docker rm -f ecr-login
+ExecStart=/bin/bash -c \
+    'source /etc/profile.d/etcdctl.sh && if [[ -n "$($AWS_ACCOUNT_ID)\" ]]; then eval $(/usr/bin/docker run --name ecr-login behance/utility aws ecr get-login --registry-ids $($AWS_ACCOUNT_ID) --region $($REGION)); fi'
```
chr0n1x: a slight nitpick, but if possible split this into more lines.

v2/fleet/aws-ecr-login.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Runs Amazon ECR Login every 6 hours
```
chr0n1x, line 3: why is this needed? maybe add some comments

v2/setup/ecr.sh
```diff
@@ -0,0 +1,15 @@
...
+
+sudo docker run --rm \
+    -v ${HOMEDIR}:/data/ behance/docker-aws-s3-downloader \
+     us-east-1 $CONTROL_TIER_S3SECURE_BUCKET .ecr
+
+# it's expected that these fields are already in the form
+# /KEY/NAMESPACE VALUE
+while read line || [[ -n "$line" ]]; do
+    etcdctl set $line
+done < ${HOMEDIR}/.ecr
```
chr0n1x, line 15: please add the content format to the [README](https://github.com/adobe-platform/mesos-systemd#files-in-s3)

chr0n1x:
Overall, looks solid. We should talk with @iderdik & @matthewdfuller about where these should live though, as I'm not quite sure what the benefits/ramifications are for doing one over the other.

v2/fleet/aws-ecr-login.timer
cdmurph32, line 3: The ECR login is only valid for 12 hours.  So I created a timer to repeat refresh the credentials.  The instances seem to reboot fairly frequently so it might not be strictly necessary.

cdmurph32:
Thanks for the timely review @chr0n1x.

I've made changes to address today's comments.  As @iderdik requested, this service will not attempt to obtain ECR credentials if /ECR/registry-account is not set.  Therefore it should be safe to run on all workers, even those that do not use ECR.

v2/fleet/aws-ecr-login.service
```diff
@@ -0,0 +1,26 @@
...
+Description=Amazon ECR Login
+After=docker.service
+Requires=docker.service
+Requires=network-online.target
+After=network-online.target
+
+[Service]
+Type=oneshot
+User=core
+Environment="AWS_ACCOUNT_ID=etcdctl get /ECR/registry-account"
```
iderdik: Can we use an ECR-specific name for this env var so it's clear what we're using it for? (i.e. ECR_AWS_ACCOUNT_ID)

matthewdfuller:
This looks great to me for now - however, it may lead to a chicken/egg situation when we begin using the ECR for all containers (which would include the container that currently downloads the ECR credentials to download containers). We'll need to discuss a way of loading ECR credentials for the first boot without pulling from Behance's Docker Hub.

iderdik:
:+1:

cdmurph32:
I just realized this service won't work for marathon.  I'll need to do some more work to have marathon use the credentials.

matthewdfuller:
Merging in per request

PR #147, opened 2016-03-07
Changes to accommodate internal IT CoreOS cluster

iderdik:
- Added new `it-hybrid` role
- Added additional env vars to FD
- Bash workaround for `etcdctl` script in capcom.sh - should probably be done for others as well


v2/fleet-local/control/flight-director@.service
```diff
@@ -2,9 +2,8 @@
 Description=Flight Director @ %i
 # it's implied that chronos & marathon require mesos
 # look here: https://github.com/behance/mesos-systemd/tree/master/v2/fleet
-After=docker.service chronos@%i.service marathon@%i.service
```
chr0n1x, line 5: ???

v2/fleet-local/control/flight-director@.service
```diff
@@ -56,5 +56,4 @@ WantedBy=multi-user.target
 
 [X-Fleet]
 Global=false
-MachineMetadata=role=control
```
chr0n1x, line 59: whoa WHAT

v2/setup/fleet.sh
```diff
@@ -9,7 +9,9 @@ fi
 
 # https://gist.github.com/skippy/d539442ada90be06459c
 # TODO: discuss anything else that would be useful here
-AZ=$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)
+if [[ -z $ZONE ]]; then
```
chr0n1x, line 12: so this env var is always set in IT envs?

iderdik, line 12: It needs to be via CloudConfig. If it isn't, it doesn't break, just hangs on this step until the metadata call times out.

v2/fleet-local/control/flight-director@.service
```diff
@@ -47,6 +45,8 @@ ExecStart=/usr/bin/sh -c "/usr/bin/docker run \
   -e GITHUB_CLIENT_ID=`etcdctl get /FD/GITHUB_CLIENT_ID` \
   -e GITHUB_CLIENT_SECRET=`etcdctl get /FD/GITHUB_CLIENT_SECRET` \
   -e AUTHORIZER_TYPE=`etcdctl get /FD/AUTHORIZER_TYPE` \
+  -e FD_MARATHON_MASTER_PROTOCOL=`etcdctl get /FD/FD_MARATHON_MASTER_PROTOCOL` \
+  -e FD_ALLOW_MARATHON_UNVERIFIED_TLS=`etcdctl get /FD/FD_ALLOW_MARATHON_UNVERIFIED_TLS` \
```
matthewdfuller, line 49: Are we allowing unverified TLS because we are allowing SSL certs that don't match the domain?

v2/setup/fleet.sh
chr0n1x, line 12: hate to be that guy - is that documented somewhere? Or going to be?
:sob:

v2/fleet-local/control/flight-director@.service
iderdik, line 49: It's configurable (off by default). Accessing the internal Mesos clusters API requires this to be set to true due to an untrusted cert:  https://www.dropbox.com/s/vyrt2xyxev41u2k/Screenshot%202016-03-07%2010.30.54.png?dl=0 .

chr0n1x, line 49: I would've assumed that this would be set in the FD setupscript, with it referencing `$ZONE`

v2/setup/fleet.sh
matthewdfuller, line 12: Was just about to ask this for the additional ENV variables it supports

chr0n1x:
- `$ZONE` is a bit ambiguous of a name, IMO.
- would it be too much to have ONLY the `init` script reference `$ZONE`?

v2/fleet-local/control/flight-director@.service
chr0n1x, line 49: The `etcd` value, that is. But I don't see that in this PR. Am I missing something?

matthewdfuller:
How are we getting AWS keys onto the instance in the first place? Also, we'll need to keep track of all the permissions it will need, which I assume will be a superset of existing tier IAM role permissions and the new Dynamo/KMS get/decrypt permissions.

v2/fleet-local/control/flight-director@.service
iderdik, line 49: The FD code itself enforces the defaults if the vars don't exist but I will explicitly add them to the FD setup.

v2/setup/fleet.sh
iderdik, line 12: The entire cloundconfig for internal IT will need to be documented. I'll include this in that

iderdik:
@matthewdfuller this issue is on the agenda to discuss at security meeting today. Currently, I pass them in via CloudConfig

matthewdfuller:
Cool, :+1:

chr0n1x:
:+1:

PR #152, opened 2016-03-09
fd-users service to watch FD etcd keys & create users

chr0n1x:
watch `/flight-director/users` for a string of `"user1 user2 user3"` usernames, use `useradd` to create them on worker instances.


matthewdfuller:
:+1:

PR #156, opened 2016-03-15
testing framework authentication between master and slave

anshupande:


v2/setup/mesos-master.sh
```diff
@@ -0,0 +1,15 @@
...
+HOMEDIR=$(eval echo "~`whoami`")
+
+AWS_CREDS=""
+if [ ! -z $AWS_ACCESS_KEY ]; then
+    AWS_CREDS=" -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY \
+     -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_KEY "
+fi
+
+sudo docker run --rm \
+    -v ${HOMEDIR}:/data/ $AWS_CREDS behance/docker-aws-s3-downloader \
```
bryanlatten, line 14: is this now in the adobe-platform repo now?

v2/setup/mesos-slave.sh
```diff
@@ -0,0 +1,15 @@
...
+HOMEDIR=$(eval echo "~`whoami`")
+
+AWS_CREDS=""
+if [ ! -z $AWS_ACCESS_KEY ]; then
+    AWS_CREDS=" -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY \
+     -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_KEY "
+fi
+
+sudo docker run --rm \
+    -v ${HOMEDIR}:/data/ $AWS_CREDS behance/docker-aws-s3-downloader \
```
bryanlatten, line 14: same comment as above

v2/setup/mesos-master.sh
anshupande, line 14: @bryanlatten If I change this for downloading one secret, I will have to change it for all secrets... I think that should be a separate PR. Let me know your views?

v2/setup/mesos-slave.sh
anshupande, line 14: we can change this in separate PR?

v2/setup/mesos-master.sh
bryanlatten, line 14: @anshupande please change it - why would we keep a reference to the behance containers?

matthewdfuller, line 14: @bryanlatten because we pull from Behance in numerous places and the change you're proposing would require an update to .dockercfg (new Docker Hub location) as well as all the images that get pulled down elsewhere. Much better to do in a separate PR.

EDIT: we can do this change in V2 as well.

matthewdfuller, line 14: Specifically, here: https://github.com/adobe-platform/mesos-systemd/blob/master/v2/setup/images.sh

anshupande:
screen shot 2016-03-22 at 12 10 44 pm

anshupande:
deployment successful and mesos_master and mesos_slave started using authentication

anshupande:
@bryanlatten @iderdik @matthewdfuller @fraczles this is ready to be merged...I need thumbs up

matthewdfuller:
We should probably test this in V3 now - is there any issue you can think of with authentication using internal routes?

anshupande:
mesos_slave while starting should have same set of secrets as are with mesos_master...If both are using same secrets we will be able to start the service..else the service will fail to start @matthewdfuller

anshupande:
so internal route should not break anything

matthewdfuller:
Okay cool, so let's merge this in and then work together on a PR for V3 support.

matthewdfuller:
:+1:

anshupande:
ok..i will add secrets for stage/prod/qe to secret server before merging

anshupande:
secrets are saved and I will merge this now

PR #165, opened 2016-04-01
Secrets and configs support

matthewdfuller:
This PR completely removes the need for the old method of using files in an S3 "secure bucket." It:
1. Re-writes v3/setup-credentials.sh to download all available secrets and configs from DynamoDB and load them into etcd.
2. Adjusts init to call setup-credentials.sh after etcd is initialized (v3 only)
3. Removes ecr.sh and parts of capcom.sh that downloaded files from S3.


init
```diff
@@ -34,6 +36,10 @@ if [[ $? = 4 ]]; then
   etcdctl set /environment/SCRIPTS-REV  $(cd $SCRIPTDIR && git branch|grep '\*'|cut -d' ' -f2 && cd -)
 fi
 
+if [[ "$VERSION" = "v3" ]]; then
```
iderdik, line 39: Why does this have to happen after setting environment?

matthewdfuller, line 39: It mainly has to happen after etcd2-setup.sh runs, which for the control tier happens on line 25. I stuck it after environment because by then, we're 100% sure etcd is running/configured.

v3/util/setup-credentials.sh
```diff
@@ -6,36 +6,60 @@ OWNER=$(whoami)
...
+    #etcdctl set $CONFIG_PATH $CONFIG_VAL &>/dev/null
+    printf "etcdctl set %s %s &>/dev/null \n" "$CONFIG_PATH" "$CONFIG_VAL" >> ${HOMEDIR}/.etcd_loader.sh
+done
+
+chmod +x ${HOMEDIR}/.etcd_loader.sh
+${HOMEDIR}/.etcd_loader.sh
+rm ${HOMEDIR}/.etcd_loader.sh
+
+# Create a dockercfg
+DOCKERCFG_CONTENTS=`sudo docker run behance/docker-aws-secrets-downloader --table $TABLE --key secrets --name DOCKERCFG --format plain`
```
iderdik, line 46: Any reason this is explicit and not a generic secret?

nBerg:
:+1:

v3/util/setup-credentials.sh
```diff
@@ -6,36 +6,60 @@ OWNER=$(whoami)
...
+echo "$DOCKERCFG_CONTENTS" > /home/${OWNER}/.dockercfg
+sudo chown -R ${OWNER}:${OWNER} /home/${OWNER}/.dockercfg
+sudo cp /home/${OWNER}/.dockercfg /root/.dockercfg
+
+# Save the Git Pull Key (must decode base64 since using "plain" option)
+GIT_PULL_KEY_CONTENTS=`sudo docker run behance/docker-aws-secrets-downloader --table $TABLE --key secrets --name GIT_PULL_KEY --format plain | base64 -d`
```
iderdik, line 52: Same question as above

v3/util/setup-credentials.sh
matthewdfuller, line 46: It's explicit because there is custom work done around it (saving it to a file, copying to root). In v2, before this update, there was also custom work done for .dockercfg.

Also, generic secrets just get loaded into etcd. I didn't want to put the contents of a file into etcd.

init
iderdik, line 39: My main issue here is that setup_credentials  will never run for other version (like itv1) unless explicitly added/

v3/util/setup-credentials.sh
matthewdfuller, line 52: Same concept, but in this case, the Git key may be in two pieces (KMS limitation, can only be 4096 chars max). If that happens, this decrypts them both and combines them into one file.

init
iderdik, line 39: Never mind - I think itv1 still uses v2 scripts

matthewdfuller, line 39: Yeah, this is v2 vs v3 not a node-role thing

iderdik:
:+1:

PR #171, opened 2016-04-18
fix bug from git issue 8182

anshupande:


v3/util/credential-slave.sh
```diff
@@ -9,7 +9,9 @@ secretslave=$(etcdctl get /frameworkauth/secret)
 
 mkdir $HOMEDIR/mesos-slave
 touch $HOMEDIR/mesos-slave/passwd
-
+sudo cp /root/.dockercfg /home/core/.dockercfg
+sudo chown core:core /home/core/.dockercfg
+chmod 0600 /home/core/mesos-slave/passwd
```
bossjones: I don't think this file exists till line 17, also, what user does this file get run as, root / core? if we are doing a `chmod` it should probably use `sudo` as well if this is not being run as root.

bossjones:
Also we should add a check to see if `/home/core/.dockercfg` is an empty file or a 0 byte file, then we should error out if that's the case. http://stackoverflow.com/questions/5920333/how-to-check-size-of-a-file has some examples of how we can try that

anshupande:
removed `cp /root/.dockercfg /home/core/.dockercfg && chown core:core /home/core/.dockercfg`

chr0n1x:
@bossjones that's for a separate PR so we can actually track down the source of the problem. This PR is so that we can guarantee that the credentials file for mesos is chmod'ed correctly and usable by the container.

matthewdfuller:
👍

PR #173, opened 2016-04-19
Adding default blank SSL setting for Capcom

matthewdfuller:
Can't use a blank `""` in Dynamo, so needs to be the default here.


bossjones:
👍

PR #174, opened 2016-04-19
splunk service units

ts3ng:
So the idea is that ops needs to declare the /splunk/SPLUNK_FORWARD_SERVER in their the dynamodb+KMS system in the secrets.json.  @chr0n1x how ops would know to inject the proper info into their secrets.json?


chr0n1x:
lol dat PR title

v3/setup/splunk.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
chr0n1x: so this file might not even be needed since we can just add it into the cluster `config`s. [Here is the example](https://github.com/adobe-platform/infrastructure/blob/master/v2/config/example-stack-all/secrets-example.json#L56). Like I was saying, if the configs are added correctly, the credential download script will automatically add them into etcd.

chr0n1x: @matthewdfuller can give you more details if needed

chr0n1x:
overall looks good!

v3/setup/splunk.sh
matthewdfuller: @chr0n1x this file is still needed if we don't want to require all of these for every launch. The values set here should be defaults, which the configs could override if needed.

v3/setup/splunk.sh
```diff
@@ -0,0 +1,7 @@
...
+# SPLUNK_FORWARD_SERVER shoudl be delcared as a config in the setup-credentials util
+
+source /etc/environment
+
+etcdctl set /splunk/SPLUNK_FORWARD_SERVER $SPLUNK_FORWARD_SERVER
```
matthewdfuller: Where is `$SPLUNK_FORWARD_SERVER` coming from? I'm assuming it is an environment variable, but did you update infrastructure to set it in user data? Or is it set somewhere else?

matthewdfuller:
Will this run all the time in tandem with Sumo Logic? How can a user decide _not_ to run one in favor of the other?

v3/setup/splunk.sh
chr0n1x: @matthewdfuller I was thinking that @ts3ng could add it via the `secrets.json` with the cluster. Have it in Dynamo and have the setup-credentials script write it to `/etc/environment`. Or do I not understand how that part works? 😕

matthewdfuller: @chr0n1x That's what I hope we can extend the secrets file to be, but atm it doesn't work for environment variables. I am going to start parsing our userdata and mesos-systemd scripts to come up with the best way to approach this. For right now, if `$SPLUNK_FORWARD_SERVER` is needed, it'd have to be set in userdata.

@ts3ng instead of doing

```
etcdctl set /splunk/SPLUNK_FORWARD_SERVER $SPLUNK_FORWARD_SERVER
```

you could make that a config in `secrets.json`. So just remove that line 6 and in `secrets.json`, add:

```
"/splunk/SPLUNK_FORWARD_SERVER": "value",
```

to the "configs" section.

matthewdfuller: Sorry if that was confusing, to summarize:
- `secrets.json` file can do configs (set via etcd), but not environment variables currently (changing soon).
- For now, keep this file with line 7 below (`etcdctl set /images/splunkforwarder "adobeplatform/docker-splunk:latest"`) since that is a default.
- Remove line 6 and require it in `secrets.json` configs section instead.

ts3ng: Thanks @chr0n1x @matthewdfuller for the guidance.  I've applied the changes to the infrastructure repo as well via pull request https://github.com/adobe-platform/infrastructure/pull/31 per @matthewdfuller instruction.

ts3ng:
@matthewdfuller It should be ran in tandem with sumologic if development/ops teams use sumlogic to manage the system.  The main driver of the splunk implementation is for CCF compliance.  The adobe security team runs there own splunk deployment that they use for security reports.  Although I don't see why teams won't move over to splunk in the future if we already ship logs to a splunk instance.  one thing to think about is concern over bandwidth usage and cost in AWS if we are shipping 2x the logs out of our systems.  I can foresee the sumologic implementation might be deprecated in the future since cloudops has been using splunk for awhile (but that's not my decsision).  For now it good we support both for now but allowing an on/off switch would be a good feature.

ts3ng:
@chr0n1x 👍 😄

chr0n1x:
@ts3ng please take a look at how we're handling [`opt`](https://github.com/adobe-platform/mesos-systemd/tree/master/v3/opt) services [here](https://github.com/adobe-platform/mesos-systemd/blob/master/init#L53) and update accordingly; I set this up last week to address @matthewdfuller's concerns, specifically to give cluster managers the ability to pick-and-choose which services they start up depending on what their teams are capable of or using.

ts3ng:
Will pick this up in this weeks sprint.  Will apply necessary changes and ping review again.

ts3ng:
@chr0n1x added splunk as a optional service per @matthewdfuller concerns.  please review and let me know if this is okay.

init
```diff
@@ -62,7 +62,7 @@ fi
 # mapped to the /opt directory of these scripts
 etcdctl get /environment/services
 if [[ $? = 4 ]]; then
-  etcdctl set /environment/services "sumologic datadog"
+  etcdctl set /environment/services "sumologic datadog splunk"
```
chr0n1x: this can be removed - users will just overwrite `/environment/services` with `"sumologic datadog splunk"` in their cluster configs if they want splunk

v3/setup/splunk.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# SPLUNK_FORWARD_SERVER is declard in infrastructure config/secrets.json
+
+source /etc/environment
+
+etcdctl set /images/splunkforwarder "adobeplatform/docker-splunk:latest"
```
chr0n1x: I would go as far as saying that this can be placed right into the systemd unit and you can just remove this file

chr0n1x:
@ts3ng two comments. Once those are addressed consider this PR as 👍ed

ts3ng:
@chr0n1x changes complete thanks for the help!

chr0n1x:
👍

chr0n1x:
@matthewdfuller I say merge?

matthewdfuller:
👍

PR #176, opened 2016-04-22
Removed Sysdig unit from v3

matthewdfuller:
Sysdig causing lots of unnecessary logs, and we'd have to create a custom container to handle framework auth, removing it for now.


bryanlatten:
👍

anshupande:
:+1:

matthewdfuller:
Lots of excitement about this one :) merging in. FYI - v3 only, so v2 clients will continue to use it until they update their infrastructure as well.

PR #178, opened 2016-04-22
init script clean up

chr0n1x:


init
```diff
@@ -1,66 +1,57 @@
...
+HOMEDIR=$(eval echo "~`whoami`")
+source /etc/environment || :
 
+# determine version, run appropriate setup-credentials script
 if [ -z "$1" ]; then
   VERSION="v2"
 else
   VERSION="$1"
 fi
+bash ${SCRIPTDIR}/$VERSION/util/setup-credentials.sh || :
```
matthewdfuller: In v3, this had to come after the setup scripts

EDIT: mainly the etcd2-setup scripts for control nodes. If it occurs before it, the etcd sets performed in the setup-credentials script are not properly set.

EDIT 2: Nevermind, I didn't realize you removed all the etcd stuff from setup-credentials and put into a different script. Please ignore.

v3/util/seed-etcd.sh
```diff
@@ -0,0 +1,30 @@
...
+    SECRET=$($DL_TABLE --key secrets --name $line)
+    SECRET_TYPE=`echo $SECRET | cut -d' ' -f2`
+    SECRET_PATH=`echo $SECRET | cut -d' ' -f3`
+    SECRET_VAL=`echo $SECRET | cut -d' ' -f4-`
+
+    if [[ "$SECRET_TYPE" != "etcd" ]]; then
+        continue
+    fi
+
+    etcdctl set -- $SECRET_PATH "$SECRET_VAL"
```
matthewdfuller, line 21: So this part won't work for some of our etcd values. I copied Iz's method of writing the commands to a file and then executing it (see L27 in old setup-credentials) to work around this.

v3/util/seed-etcd.sh
```diff
@@ -0,0 +1,30 @@
...
+done
+
+echo "$AV_CONFIGS" | while read line ; do
+    CONFIG=$($DL_TABLE --key configs --name $line)
+    CONFIG_PATH=`echo $CONFIG | cut -d' ' -f1`
+    CONFIG_VAL=`echo $CONFIG | cut -d' ' -f2-`
+
+    etcdctl set $CONFIG_PATH "$CONFIG_VAL"
```
matthewdfuller, line 29: Another location where I was using a file, but if we don't use the file, we'll still need at least a `--` here too.

chr0n1x, line 29: ugggh I _really_ want to avoid doing that

matthewdfuller, line 29: We can, we just need to find a way to pass every possible character that might be in an etcd set. For example: `etcdctl set -- AIRLOCK_PUBLIC_KEY ----PUBLIC KEY----\n...` failed spectacularly last time I tried. Same when the secret had a number of different special characters.

chr0n1x, line 29: I'll give it a shot

PR #180, opened 2016-04-27
v3: pull RDSPASSWORD from Dynamo & populate

chr0n1x:


v3/util/setup-credentials.sh
```diff
@@ -34,6 +34,13 @@ if [ -f ${HOMEDIR}/.ssh/id_rsa ]; then
     cat ${HOMEDIR}/.ssh/id_rsa.pub >> ${HOMEDIR}/.ssh/authorized_keys
 fi
 
+# Save the RDS password to environment variable in control tier
+# TODO: :(
+if [[ "$NODE_ROLE" = "control" && $AV_SECRETS == *"RDSPASSWORD"* ]]; then
+    RDSPASSWORD=`sudo docker run behance/docker-aws-secrets-downloader --table $TABLE --key secrets --name RDSPASSWORD --format plain`
+    sudo echo "RDSPASSWORD=$RDSPASSWORD" >> /etc/environment
```
chr0n1x: @matthewdfuller so this doesn't work unless the bootstrap script is run as `sudo`. Not sure how we want to handle this.

chr0n1x:
@matthewdfuller @iderdik I don't love it, and this will change when we start to clean the FD keys. Thoughts?

matthewdfuller:
It gets the job done for now. I like this better than my previous idea of having the setup script inject a line at the end of /etc/environment. I think that file should remain a read-only exact copy of what was written at boot.

matthewdfuller:
👍

chr0n1x:
ok - I'll A/B test this right after https://github.com/adobe-platform/mesos-systemd/pull/182

PR #182, opened 2016-04-28
v3/capcom: reorganize & scope etcd values

chr0n1x:
Resolves https://github.com/adobe-community/issues/issues/8581

### DIFF
--- a/v3/setup/capcom.sh
+++ b/v3/setup/capcom.sh
@@ -7,26 +7,24 @@ if [ "${NODE_ROLE}" != "control" ]  && [ "${NODE_ROLE}" != "it-hybrid" ]; then
     exit 0
 fi
 
-if [ "$(etcdctl get capcom-bootstrapped)" == "true" ]; then
+if [ "$(etcdctl get /capcom/config/bootstrapped)" == "true" ]; then
     exit 0
 fi
 
-
-etcdctl set capcom-bootstrapped true
-
-etcdctl setdir CP
-etcdctl set /CP/CP_APPLICATIONS '[]'
-etcdctl set /CP/CP_HOST 127.0.0.1
-etcdctl set /CP/CP_DB_PATH ./capcom.db
-etcdctl set /CP/CP_KV_STORE_SERVER_ADDRESS  http://$CAPCOM_KV_ENDPOINT
-etcdctl set /CP/CP_KV_TTL 10
-etcdctl set /CP/CP_LOG_LEVEL "$CAPCOM_LOG_LEVEL"
-etcdctl set /CP/CP_LOG_LOCATION "$CAPCOM_LOG_LOCATION"
-etcdctl set /CP/CP_PORT 2002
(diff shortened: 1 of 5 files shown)


v3/setup/capcom.sh
```diff
@@ -7,26 +7,24 @@ if [ "${NODE_ROLE}" != "control" ]  && [ "${NODE_ROLE}" != "it-hybrid" ]; then
     exit 0
 fi
 
-if [ "$(etcdctl get capcom-bootstrapped)" == "true" ]; then
+if [ "$(etcdctl get /capcom/config/bootstrapped)" == "true" ]; then
```
iderdik: Is this really a config? Seems like it might be more of a "cluster-state" value.

chr0n1x: hm good point. I was reading it more as `has the capcom config been bootstrapped?`. Hence the order & namespace.

Do you think it should look like something `/cluster-state/capcom/bootstrapped`?

matthewdfuller:
👍

v3/setup/capcom.sh
iderdik: I think that would be cleaner - everything in `config` is an actual config value. Not sure  I love the word cluster-state but something along those lines.

matthewdfuller: What about a `/state` path for each service? Like `/capcom/state/bootstrapped` or `/flight-director/state/xyz`, etc.

v3/setup/capcom.sh
```diff
@@ -7,26 +7,24 @@ if [ "${NODE_ROLE}" != "control" ]  && [ "${NODE_ROLE}" != "it-hybrid" ]; then
     exit 0
 fi
 
-if [ "$(etcdctl get capcom-bootstrapped)" == "true" ]; then
-    exit 0
+etcdctl get /capcom/config/proxy
```
chr0n1x: @iderdik @matthewdfuller how about this instead?

matthewdfuller: That works - just rely on it being set or not to determine if it was bootstrapped. As long as this file runs _before_ the etcd-seed script. If not, the user could set this value in the config file and then this script would think all the values were already populated.

iderdik: I don't love this because if I have to rerun the config load, I currently set the `bootstrapped` flag to false and then run it. This will mean I have to delete all the keys.

matthewdfuller: We could just create an entirely new `/flags` top-level dir: `/flags/capcom-bootstrapped` etc

chr0n1x: hm ok. Then a parent namespace it is.
What about:

```
etcdctl set /bootstrap.service/capcom true
```

The semantics and namespace implies that it belongs to the bootstrap.service unit. If we ever want to use the `/bootstrap.service/capcom` key for anything else, we can just extend it.

iderdik: I like the bootstrap namespace.

PR #187, opened 2016-05-03
klam-ssh support for CoreOS

eadasiak:
This adds support for klam-ssh on CoreOS.

The klam-ssh.sh script in v3/setup installs the pre-reqs:
1.  A replacement for /etc/nsswitch.conf that includes the libnss_ato library method for passwd, shadow, and group
2.  The klam-ssh.conf file in /opt/klam/lib
3.  The libnss_ato shared library in /opt/klam/lib64, and the ld.so.conf.d config for that file path
4.  Reconfigures the linker cache
5.  Adds the klamfed user (all klam-ssh users are squashed to this user)
6.  Adds sudo support for klamfed
7.  Moves supporting scripts into place in /opt/klam/lib
8.  Appends config to the sshd_config file for the AuthorizedKeysCommand

The authorizedkeys_command.sh script runs the klam-ssh docker container to get the user's ssh key from S3

The downloadS3.sh script runs the klam-ssh docker container to get the list of authorized klam-ssh users via S3.

Let me know if you have any questions.  Next stop -- the bastion host!


v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,137 @@
...
+ROLE_NAME="$(etcdctl get /klam-ssh/ROLE_NAME)"
+ENCRYPTION_ID="$(etcdctl get /klam-ssh/ENCRYPTION_ID)"
+ENCRYPTION_KEY="$(etcdctl get /klam-ssh/ENCRYPTION_KEY)"
+KEY_LOCATION_PREFIX="$(etcdctl get /klam-ssh/KEY_LOCATION_PREFIX)"
+IMAGE="$(etcdctl get /images/klam-ssh)"
+
+
+if [[ $REGION == "eu-west-1" ]]; then
+    KEY_LOCATION="-ew1"
+elif [[ $REGION == "ap-northeast-1" ]]; then
```
bryanlatten: should we make this list accessible to the widest swath of regions, or is unnecessary?

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,137 @@
+#!/bin/bash
```
bryanlatten: should we slap an `-e` on there to exit if any command fails?

v3/profile.d/klam.sh
```diff
@@ -0,0 +1,2 @@
+KLAM_USER=$(who -m | awk '{print $1}')
+PROMPT_COMMAND='RETRN_VAL=$?;logger -p local6.debug "$KLAM_USER [$$]: $(history 1 | sed "s/^[ ]*[0-9]\+[ ]*//") [$RETRN_VAL]"'
```
chr0n1x: this is some...voodoo magic.

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,137 @@
...
+
+AZ=$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)
+REGION=${AZ::-1}
+ROLE_NAME="$(etcdctl get /klam-ssh/ROLE_NAME)"
+ENCRYPTION_ID="$(etcdctl get /klam-ssh/ENCRYPTION_ID)"
```
chr0n1x: lowercase & dashes for all of these

v3/setup/klam-ssh.sh
chr0n1x: agreed. I'm also starting to lean towards `-xe` due to the amount of stuff going on in here

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,137 @@
...
+elif [[ $REGION == "us-west-1" ]]; then
+    KEY_LOCATION="-uw1"
+elif [[ $REGION == "us-west-2" ]]; then
+    KEY_LOCATION="-uw2"
+else
+    echo "An incorrect region value specified"
+    exit 1
+fi
+
+# create nsswitch.conf
```
chr0n1x: this is a huge nitpick but there _has_ to be a better way to do this

v3/setup/klam-ssh.sh
matthewdfuller: Also, put these in `/klam-ssh/config/your-value-here`

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,137 @@
...
+echo "services:   files usrfiles" >> /home/core/nsswitch.conf
+echo "protocols:  files usrfiles" >> /home/core/nsswitch.conf
+echo "rpc:        files usrfiles" >> /home/core/nsswitch.conf
+echo -e "\n" >> /home/core/nsswitch.conf
+echo "ethers:     files" >> /home/core/nsswitch.conf
+echo "netmasks:   files" >> /home/core/nsswitch.conf
+echo "netgroup:   nisplus" >> /home/core/nsswitch.conf
+echo "bootparams: files" >> /home/core/nsswitch.conf
+echo "automount:  files nisplus" >> /home/core/nsswitch.conf
+echo "aliases:    files nisplus" >> /home/core/nsswitch.conf
```
matthewdfuller: Another stylistic nitpick, but can't we do something like:

```
cat <<EOT >> /home/core/nsswitch.conf
passwd: files usrfiles ato
shadow: files usrfiles ato
...
EOT
```

Just makes things easier to read / cleaner.

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,137 @@
...
+# create klam-ssh.conf
+echo "{" > /home/core/klam-ssh.conf
+echo "    \"key_location\": \"${KEY_LOCATION_PREFIX}${KEY_LOCATION}\"," >> /home/core/klam-ssh.conf
+echo "    \"role_name\": \"${ROLE_NAME}\"," >> /home/core/klam-ssh.conf
+echo "    \"encryption_id\": \"${ENCRYPTION_ID}\"," >> /home/core/klam-ssh.conf
+echo "    \"encryption_key\": \"${ENCRYPTION_KEY}\"," >> /home/core/klam-ssh.conf
+echo "    \"resource_location\": \"amazon\"," >> /home/core/klam-ssh.conf
+echo "    \"time_skew\": \"permissive\"," >> /home/core/klam-ssh.conf
+echo "    \"s3_region\": \"${REGION}\"" >> /home/core/klam-ssh.conf
+echo "}" >> /home/core/klam-ssh.conf
```
matthewdfuller: Same here: `cat <<EOT >> /home/core/klam-ssh.conf`

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,137 @@
...
+mv /home/core/mesos-systemd/v3/util/authorizedkeys_command.sh /opt/klam/lib
+
+# Change ownership of downloadS3
+chown root:root /home/core/mesos-systemd/v3/util/downloadS3.sh
+
+# Relocate downloadS3.sh
+mv /home/core/mesos-systemd/v3/util/downloadS3.sh /opt/klam/lib
+
+# Restart SSHD
+systemctl restart sshd.service
```
matthewdfuller: Is this going to conflict with the existing service which restarts sshd? Can we reduce how many times we are restarting it by making sure all the conf changes are made first, then restarting it once?

v3/util/authorizedkeys_command.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+ENCRYPTION_ID=$(etcdctl get /klam-ssh/ENCRYPTION_ID)
+ENCRYPTION_KEY=$(etcdctl get /klam-ssh/ENCRYPTION_KEY)
+ROLE_NAME=$(etcdctl get /klam-ssh/ROLE_NAME)
+KEY_LOCATION_PREFIX=$(etcdctl get /klam-ssh/KEY_LOCATION_PREFIX)
```
matthewdfuller: Make sure to update these to `/klam-ssh/config/your-value-here` as well.

matthewdfuller:
This is doing some crazy stuff, but if it works...

v3/setup/klam-ssh.sh
chr0n1x: 👍

v3/setup/klam-ssh.sh
chr0n1x: I don't _think_ so? It shouldn't just because this will run serially. @eadasiak can you test and confirm?

matthewdfuller: Conflict was probably the wrong word; I'm thinking more of the fact that ssh is being restarted twice. We already have to wait about five minutes before we can SSH in to an instance, this might make that longer.

matthewdfuller:
@eadasiak once this is merged, will regular SSH logins stop working?

v3/util/authorizedkeys_command.sh
```diff
@@ -0,0 +1,12 @@
...
+IMAGE=$(etcdctl get /images/klam-ssh)
```
matthewdfuller: `'images/klam-ssh` should be set (as a default) in `v3/setup/klam-ssh.sh`

v3/setup/klam-ssh.sh
chr0n1x: The restart is actually quite fast though. We see the daemon restart when fleet is restarted and it's near instantaneous to establish a new connection

v3/setup/klam-ssh.sh
eadasiak: Unfortunately, I was limited to the regions where klam-ssh is configured.  Version 2.0 of klam-ssh will be more widely available, and will get addressed with its release.

v3/setup/klam-ssh.sh
eadasiak: Good point!  Made that change.

v3/profile.d/klam.sh
eadasiak: Got rid of this and made a change to the global command prompt instead to use the `who -m` voodoo instead.  This wasn't working correctly under CoreOS as it was, so away it goes!

v3/setup/klam-ssh.sh
eadasiak: OK, I fixed these up.  There will need to be an update to infrastructure as well so that the example configs have the values corrected.

v3/setup/klam-ssh.sh
eadasiak: If "unhappy"; then
  how about a case statement?
elif "unsatisfied"; then
  how about a case statement?
else
  you get a case statement
  exit 0

v3/setup/klam-ssh.sh
eadasiak: You got it!

v3/setup/klam-ssh.sh
eadasiak: I wasn't able to find where the other sshd restart happens.  This particular restart is really fast, but if the sequence of events is such that the setup scripts occur before the other service restart, I can strike this one.

v3/util/authorizedkeys_command.sh
eadasiak: Done!

v3/util/authorizedkeys_command.sh
eadasiak: Ah, I see your point.  I was relying on the configs from the infrastructure repo to do this instead.  I've set a default value in v3/setup/klam-ssh.sh

eadasiak:
@matthewdfuller: As long as the infrastructure stack still provides a mechanism to provide an ssh key, then the core user will still be able to log in.

v3/util/authorizedkeys_command.sh
eadasiak: Associated infrastructure PR https://github.com/adobe-platform/infrastructure/pull/71

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,144 @@
...
+protocols:  files usrfiles
+rpc:        files usrfiles
+
+ethers:     files
+netmasks:   files
+netgroup:   nisplus
+bootparams: files
+automount:  files nisplus
+aliases:    files nisplus
+EOT
```
matthewdfuller: So much cleaner! :)

v3/setup/klam-ssh.sh
matthewdfuller: I'm okay with leaving this here; was just thinking about optimizing it, but if you say it's fast, LGTM

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,144 @@
+#!/bin/bash -xe
+
+etcdctl set /images/klam-ssh  "adobecloudops/klam-ssh:latest"
+
+AZ=$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)
+REGION=${AZ::-1}
+ROLE_NAME="$(etcdctl get /klam-ssh/config/role-name)"
```
matthewdfuller: If the KLAM Role is created in the infrastructure stack, it should be added as an environment variable here: https://github.com/adobe-platform/infrastructure/blob/master/v2/src/userdata/templates/write_files/environment/shared.yml

Otherwise, we will have to have users pull their KLAM role out of the infrastructure stack and insert it into secrets.

matthewdfuller:
👍

v3/setup/klam-ssh.sh
```diff
@@ -0,0 +1,144 @@
...
+
+#  update /etc/ssh/sshd_config
+cp /etc/ssh/sshd_config sshd_config
+echo 'AuthorizedKeysCommand /opt/klam/lib/authorizedkeys_command.sh' >> sshd_config
+echo 'AuthorizedKeysCommandUser root' >> sshd_config
+mv -f sshd_config /etc/ssh/sshd_config
+cat /etc/ssh/sshd_config
+
+# Change ownership of authorizedkeys_command
+chown root:root /home/core/mesos-systemd/v3/util/authorizedkeys_command.sh
```
chr0n1x: huh...does this actually work? And why is this needed?

eadasiak:
@chr0n1x @matthewdfuller - I found a few bugs with the original implementation, and have checked in the fixes.  In particular, the klam-ssh setup script being called before etcd on the hosts was ready.  To fix that, I've moved the script from /setup to /util, and modified the `init` script to call it explicitly.  I've also added a helper script that a user can call to configure their local ssh config so that they connect to internal hosts through the bastion with klam support.

PR #192, opened 2016-05-05
V3: re-working mesos & framework credential setup

chr0n1x:
- renamed `framework` key to `mesos` for clarity
- made `framework` & `marathon` both have `config` subdirs
- simplified & consolidated credential file logic


v3/setup/mesos-credentials.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+# ensure that we have sane defaults for credentials
+etcdctl get /mesos/config/username
+if [[ $? = 4 ]]; then
+    etcdctl set /mesos/config/username ethos
+fi
+etcdctl get /mesos/config/password
+if [[ $? = 4 ]]; then
+    etcdctl set /mesos/config/password password
```
matthewdfuller: Not sure how I feel about this - on one hand the stack will come up if a user forgets to pass in a password. On the other hand, we're hardcoding insecure defaults.

matthewdfuller:
Other than the one security comment, 👍

v3/setup/mesos-credentials.sh
chr0n1x: I guess that I could just exit the script with a very "loud" warning? This will prevent the credential files from being written and the changes that I added in the `ExecStartPre` steps of the units will result in failures.

anshupande:
:+1:

v3/setup/mesos-credentials.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+source /etc/environment
+if [ "${NODE_ROLE}" != "control" ] && [ "${NODE_ROLE}" != "worker"]; then
+    exit 0
+fi
+
+etcdctl get /mesos/config/password
+if [[ $? = 4 ]]; then
+    echo "etcd /mesos/config/password NOT SET, NOT CONTINUING"
```
chr0n1x, line 11: @matthewdfuller better?

matthewdfuller, line 11: Much!

v3/fleet-local/worker/mesos-slave@.service
```diff
@@ -19,10 +19,10 @@ ExecStartPre=/usr/bin/systemctl is-active update-os.service
 # condition / assertion options here for some reason do not prevent the
 # ExecStart from spinning up and mounting / creating .dockercfg/ as a dir
 ExecStartPre=/usr/bin/bash -c "if [ ! -f /home/core/.dockercfg ]; then exit 1; fi"
+ExecStartPre=/usr/bin/bash -c "if [ ! -f /opt/mesos/credentials ]; then exit 1; fi"
```
chr0n1x, line 22: @matthewdfuller in the case where the credentials are bad (i.e.: missing `/mesos/config/password`), this file will not exist and the cluster just wont spin up.

PR #194, opened 2016-05-06
v3/setup/images: move bootstrap flags to bootstrap.service

chr0n1x:
I believe this is all that's left for the keys

### DIFF
--- a/v3/setup/images-control.sh
+++ b/v3/setup/images-control.sh
@@ -6,11 +6,11 @@ if [ "${NODE_ROLE}" != "control" ]; then
     exit 0
 fi
 
-if [ "$(etcdctl get images-control-bootstrapped)" == "true" ]; then
+if [ "$(etcdctl get /bootstrap.service/images-control-bootstrapped)" == "true" ]; then
     echo "control-tier images already bootstrapped, skipping"
     exit 0
 fi
-etcdctl set images-control-bootstrapped true
+etcdctl set /bootstrap.service/images-control-bootstrapped true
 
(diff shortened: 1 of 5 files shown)


v3/setup/images-control.sh
```diff
@@ -6,11 +6,11 @@ if [ "${NODE_ROLE}" != "control" ]; then
     exit 0
 fi
 
-if [ "$(etcdctl get images-control-bootstrapped)" == "true" ]; then
+if [ "$(etcdctl get /bootstrap.service/images-control-bootstrapped)" == "true" ]; then
```
matthewdfuller, line 9: Super minor nitpick, but in other places we've used `/bootstrap.service/flight-director` (https://github.com/adobe-platform/mesos-systemd/blob/master/v3/setup/flight-director.sh#L8). Would it make more sense to call this `bootstrap.service/images-control` or even `bootstrap.service/images/control`?

matthewdfuller:
👍

v3/setup/images-control.sh
chr0n1x, line 9: hmmm not totally sure. I don't see where we would use `bootstrap.service/images/control` for though.

chr0n1x, line 9: at that point, I guess I could move the `/images` dir into `bootstrap.service`. But having to type out `bootstrap.service/images/control/flight-director` is a lot. I don't know

matthewdfuller, line 9: The above is okay then, my comment was more that `/bootstrap.service` already implies that whatever follows (images-control-bootstrapped) is being bootstrapped. I.e., we don't call it `/bootstrap.service/flight-director-bootstrapped`

chr0n1x, line 9: eh if it becomes an issue at least the mess will be contained :P

PR #198, opened 2016-05-10
Adding RSA private key generation

matthewdfuller:
- Ensuring `~/.ssh/id_rsa` exists

### DIFF
--- a/v3/util/setup-credentials.sh
+++ b/v3/util/setup-credentials.sh
@@ -27,6 +27,8 @@ sudo cp /home/${OWNER}/.dockercfg /root/.dockercfg
 if [ -f ${HOMEDIR}/.ssh/id_rsa ]; then
     ssh-keygen -f ${HOMEDIR}/.ssh/id_rsa -y > ${HOMEDIR}/.ssh/id_rsa.pub
     cat ${HOMEDIR}/.ssh/id_rsa.pub >> ${HOMEDIR}/.ssh/authorized_keys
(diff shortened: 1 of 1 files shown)


v3/util/setup-credentials.sh
```diff
@@ -27,6 +27,8 @@ sudo cp /home/${OWNER}/.dockercfg /root/.dockercfg
 if [ -f ${HOMEDIR}/.ssh/id_rsa ]; then
     ssh-keygen -f ${HOMEDIR}/.ssh/id_rsa -y > ${HOMEDIR}/.ssh/id_rsa.pub
     cat ${HOMEDIR}/.ssh/id_rsa.pub >> ${HOMEDIR}/.ssh/authorized_keys
+    # Actually generate a private key
+    ssh-keygen -f ${HOMEDIR}/.ssh/id_rsa -t rsa -N ''
```
chr0n1x: shouldn't this be before the public key gets generated?

v3/util/setup-credentials.sh
```diff
@@ -27,6 +27,8 @@ sudo cp /home/${OWNER}/.dockercfg /root/.dockercfg
 if [ -f ${HOMEDIR}/.ssh/id_rsa ]; then
```
chr0n1x: the `if` statement can go away now too

chr0n1x:
👍 thanks for doing this!

PR #199, opened 2016-05-10
init: hard-lock script version

chr0n1x:
We're essentially feature complete (bare-bones deployment pipeline) so this is now needed since the majority of our work has been dealing with organiziation of data & service add-ons. Doing this will hard-lock clusters to a particular version of the scripts, which is good, and maintainers will force maintainers to  review changes and upgrade scripts manually (preferably via control-jenkins).

### DIFF
--- a/init
+++ b/init
@@ -53,9 +53,17 @@ etcdctl get /environment/SCRIPTS-FORK
 if [[ $? = 4 ]]; then
   # 4 == 404 - key not found
   # case where a node is joining a pre-existing cluster
-  SCRIPTS_REV=$(cd $SCRIPTDIR && git branch|grep '\*'|cut -d' ' -f2 && cd -)
+  SCRIPTS_REV=$(cd $SCRIPTDIR && git rev-parse HEAD)
   etcdctl set /environment/SCRIPTS-FORK adobe-platform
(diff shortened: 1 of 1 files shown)


adobejmong:
@chr0n1x if the intent is to lock it a specific sha,  shouldn't the command be something like `git reset --hard <sha>`? 

Also what if we just tag/create releases for mesos-systemd just like what we used to do in infrastructure? Will that be too much of an overhead?

Its a lot easier for me to work with clients/help others if I can just reference tag x.y.z for example. What do you think?

adobejmong:
@chr0n1x thanks for the explanation in slack of locking it to the infrastructure version so 👍 from me

I still would love to tag releases as we did before if at all possible since its a lot easier when talking to clients and helping them

chr0n1x:
@adobejmong @matthewdfuller this is still a thing that needs to be merged so that people who hard-lock to a version of these scripts in their bootstrap units will also have their locked versions respected by automations that rolling deploy other units. We ok with merging this?

chr0n1x:
ping @adobe-platform/platform-core

adobejmong:
👍

matthewdfuller:
👍

PR #201, opened 2016-05-12
Running mesos-credentials script after seed-etcd

matthewdfuller:
- Fixed issue where some instances don't come up as healthy due to missing `/mesos/config` etcd values.


chr0n1x:
👍

PR #203, opened 2016-05-12
init: retry submitting fleet units

chr0n1x:


adobejmong:
👍 

I feel confident with this fix and was able to spin up a cluster without issue

matthewdfuller:
👍

PR #205, opened 2016-05-13
Adding Sumo Logic Hostname Source

matthewdfuller:
- Added `SUMO_HOSTNAME` environment variable to sumo service files.
- When passed to Sumo (see https://github.com/behance/docker-sumologic/pull/2/files), will be used to set the `sourceHost` to the IP of the the host rather than the IP of the container.


chr0n1x:
👍

PR #206, opened 2016-05-16
make changes from anshu's fork for etcdctl authentication

anshupande:


chr0n1x:
@anshupande please squash your commits

v3/setup/etcdctl-auth.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
matthewdfuller: Change this to `#!/bin/bash -x` for logging purposes

v3/setup/etcdctl-auth.sh
```diff
@@ -0,0 +1,53 @@
...
+if [[ ! -d $CRED_DIR ]]; then
+    sudo mkdir $CRED_DIR -p
+fi
+
+# add root-user
+sudo echo '{"user":"'${ROOT_USERNAME}'", "password":"'${ROOT_PASSWORD}'"}' > $CRED_DIR/root-user.json
+curl -L http://127.0.0.1:2379/v2/auth/users/${ROOT_USERNAME} -XPUT -d "@$CRED_DIR/root-user.json"
+
+# add read-user
+sudo echo '{"user":"'${READ_USERNAME}'", "password":"'${READ_PASSWORD}'"}' > $CRED_DIR/read-user.json
```
adobejmong: @anshupande I think we're better off using a heredoc instead of using echo like this as this could end up being error prone (you missing a quote etc)

so the line above can be rewritten as something like

```
sudo cat << EOF > $CRED_DIR/read-user.json
{
  "user": "${READ_USERNAME}",
  "password": "${READ_PASSWORD}"
}
EOF
```

It's also a lot easier to read/troubleshoot

chr0n1x: agreed

v3/setup/etcdctl-auth.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+source /etc/environment
```
chr0n1x: I don't think you need this here

v3/setup/etcdctl-auth.sh
```diff
@@ -0,0 +1,53 @@
...
+etcdctl role add read-write
+etcdctl role grant read-write -path '/*' -readwrite
+
+# Give read-user read access
+sudo echo '{"user": "'${READ_USERNAME}'", "grant": ["read-only"]}' > $CRED_DIR/read-only.json
+curl -L http://127.0.0.1:2379/v2/auth/users/${READ_USERNAME} -XPUT -d "@$CRED_DIR/read-only.json"
+
+# Give read-write write access
+sudo echo '{"user": "'${WRITE_USERNAME}'", "grant": ["read-write"]}' > $CRED_DIR/read-write.json
+curl -L http://127.0.0.1:2379/v2/auth/users/${WRITE_USERNAME} -XPUT -d "@$CRED_DIR/read-write.json"
```
adobejmong, line 60: @anshupande I also just realized this but you do what appears to be similar calls for all intents and purposes, have you thought about just wrapping the whole thing in a function and calling that? Are the files in `$CRED_DIR` ever re-used? Or they are loaded and then discarded? If discarded, I don't see a corresponding removal of files in the PR

The other remark I will say is maybe you create all of the files up front first and then call storing them later e.g. group all the creation of the files on top and then run your curl calls at the bottom.

```
... define all the files here ...

... run all the curl commands here ...
```

Just my 2 cents

v3/setup/etcdctl-auth.sh
anshupande: Yes , I can reformat it and a function

v3/setup/etcdctl-auth.sh
anshupande, line 60: @adobejmong yes we do not need $CRED_DIR , but we will need root username and password after we delete the guest role. In this PR I will delete the $CRED_DIR

v3/setup/etcdctl-auth.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Enable authentication
+curl -L http://127.0.0.1:2379/v2/auth/enable -XPUT
+
+sudo rm -rf $CRED_DIR
```
adobejmong, line 65: @anshupande  just being paranoid here but maybe change line 52 to be something like

```
[[ -d "$CRED_DIR" ]] && sudo rm -rf "${CRED_DIR}"
```

chr0n1x, line 65: the `-f` makes it ok I think.

v3/setup/etcdctl-auth.sh
anshupande, line 60: Ok Added a function to add users

anshupande:
I will delete comments after testing function that adds user tomorrow @adobejmong @chr0n1x

anshupande:
@chr0n1x @adobejmong I have tested this again...and this works...let me know if we can merge?

chr0n1x:
@anshupande you still have not squashed your commits

PR #207, opened 2016-05-16
Bugfix add marathon authentication to docker-dd-agent-mesos service

digitalyuki:
With the new marathon authentication added to mesos-systemd v3, the datadog agent running on the control node broke. This will add in the marathon username and password values from etcd into the docker container environmental variables.

https://github.com/behance/docker-dd-agent-mesos/pull/2


matthewdfuller:
👍

PR #210, opened 2016-05-18
v3/util-units/etcd2-setup: sleep until all members are healthy

chr0n1x:
Resolves https://github.com/adobe-community/issues/issues/8984


matthewdfuller:
Yay! I knew there was something we forgot last week, and this was it.
👍

chr0n1x:
How I tested:
1. SSH'ed into a single node
2. `systemctl stop etcd2`
3. watch another node's running instance of this script (it kept sleeping)
4. `systemctl start etcd2` on the instance that I stopped it
5. watched the other instance continue/break out of the loop

PR #212, opened 2016-05-19
Configurable Cgroup Sizing

matthewdfuller:
Proxy reloads can use a pre-defined maximum allotment of memory. When using c4.2xlarge instances, this was hard-coded as `12GB`. However, now that numerous instance sizes are supported, this value should be adjusted to ~75% of memory. This code dynamically calculates this value.


matthewdfuller:
Tested by re-submitting Proxy service and observing this process (on a machine with 4GB of memory):

```
docker run --name proxy --net=host -m 2825M -v /etc/nginx:/etc/nginx nginx:1.9.5 nginx -g daemon off;
```

chr0n1x:
👍

anshupande:
👍

PR #213, opened 2016-05-19
add --security-opt=no-new-privileges to chronos.service

anshupande:
A process can set the `no_new_priv` bit in the kernel. It persists across fork, clone and execve. The `no_new_priv` bit ensures that the process or its children processes do not gain any additional privileges via suid or sgid bits. This way a lot of dangerous operations become a lot less dangerous because there is no possibility of subverting privileged binaries. `no_new_priv` prevents LSMs like SELinux from transitioning to process labels that have access not allowed to the current process.


anshupande:
do not merge. still testing

anshupande:
I checked by opening a issue with Docker and here is the response : `This feature was not in docker 1.9, I think it was only added in 1.11.` https://github.com/docker/docker/issues/22862

anshupande:
Closing this and will rework on this once we are on 1.11

PR #214, opened 2016-05-19
init: remove self-update/recurse

chr0n1x:
We're going to have to figure out a way to do this eventually though :(


matthewdfuller:
👍

PR #215, opened 2016-05-19
RSA Key Check before Create

matthewdfuller:
- In preparation for potentially looping the init script, now check for the id_rsa key before creating it. If not, the command forces an interactive prompt (press Y to overwrite).


chr0n1x:
👍

anshupande:
👍

PR #216, opened 2016-05-19
Fully Qualified Docker URLs

matthewdfuller:
- Added the default `index.docker.io` before each `docker pull org/image` command. No functionality change now, but when ECR is integrated, we will need to distinguish between Docker registries.


chr0n1x:
innnnnntere-table-setting

chr0n1x:
👍

anshupande:
👍

v3/fleet-local/it-hybrid/memcached.service
```diff
@@ -10,7 +10,7 @@ Restart=always
 RestartSec=5
 TimeoutStartSec=0
 
-ExecStartPre=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && docker pull borja/docker-memcached"
+ExecStartPre=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && docker pull index.docker.io/borja/docker-memcached"
```
bryanlatten, line 13: should we also lock to a version? seems like a silly thing to allow to be unreliable

matthewdfuller, line 13: @bryanlatten Technically, yes - but based on discussions we've had so far, we will always use "latest" in mesos-systemd. If a lock is needed (up to the launching user's discretion), it can be defined in the configs section of the secrets.json file.

PR #219, opened 2016-05-20
v*/util-units/zk-health: update for exhibitor basic auth

chr0n1x:


nBerg:
:+1:

matthewdfuller:
Are you sure we want to update v2 too? If this is a critical bugfix, then I say go for it, but otherwise, this will impact all existing running clusters.

anshupande:
I also thought that we do not update v2 now with new features. But looks like this is to fix old deployment ? Otherwise 👍

matthewdfuller:
👍

PR #221, opened 2016-05-23
fleet & util: fd-user creation scripts

chr0n1x:
watch /flight-director/users for a string of "user1 user2 user3" usernames, use useradd to create them on worker instances.

### DIFF
--- a/v3/fleet/fd-users.service
+++ b/v3/fleet/fd-users.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=Create application users based on flight-director keys
+After=etcd2.service
+
+[Service]
+User=core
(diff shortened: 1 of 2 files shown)


v3/fleet/fd-users.service
```diff
@@ -0,0 +1,18 @@
...
+Description=Create application users based on flight-director keys
+After=etcd2.service
+
+[Service]
+User=core
+Restart=always
+TimeoutStartSec=0
+RestartSec=10
+
+ExecStart=/home/core/mesos-systemd/v2/util/fd-users.sh
```
chr0n1x, line 11: note to self: update this path

PR #223, opened 2016-05-23
Switch v2 datadog-control.service to use docker-dd-agent-mesos:v2

digitalyuki:
This is a change of mesos-systemd/v2 (Portfolio) to fix the version of docker-dd-agent-mesos on older clusters (Porfolio) to a fixed 'v2' branch, and then allow for updates and upgrades for docker-dd-agent-mesos agent on the latest ethos mesos-systemd/v3 to happen without breaking backwards compatibility on our older system.


chr0n1x:
:+1:

PR #226, opened 2016-05-26
docker logrotation

f4tq:
Docker logs need to be rotated by the container instances themselves when using the default json logger because the container maintains an open file handle to the backing log file.  From Coreos, this file can be found in `/var/lib/docker/containers/*/*.log`.  To remove them from Cores though causes overlay and devicemapper corruption (unable to remove layers, images and or instances).
In summary,
- All `docker run` commands have been provisioned with max-file & max-size.  The values are pulled from etcd.
- A default etcd value is set  in `init` if the user doesn't set one via secrets.json
- The external logrotate unit is removed.  
- The docker cleanup script is changed to use the docker client to remove dead containers and untagged layers i.e. those that show as '<none>' when `docker images` is run.  
- Setting secrets.json `type=etcd` and key of `/docker/config/logs-max-size` and `/docker/config/logs-max-file` override the default of `max-size=10m` and `max-file=10` i.e. no more than 10 rotated logs and of no more than 10MB in size.

Finally, the related ticket is https://github.com/adobe-community/issues/issues/9054

### DIFF
--- a/v3/fleet-local/worker/mesos-slave@.service
+++ b/v3/fleet-local/worker/mesos-slave@.service
@@ -30,6 +30,7 @@ ExecStart=/usr/bin/bash -c "source /etc/profile.d/etcdctl.sh && \
     --net=host \
     --pid=host \
     --privileged \
+    -p 5051:5051 \
     -v /home/core/.dockercfg:/root/.dockercfg:ro \
     -v /sys:/sys \
     -v /usr/bin/docker:/usr/bin/docker:ro \
--- a/v3/opt/datadog/datadog-control.service
+++ b/v3/opt/datadog/datadog-control.service
@@ -18,8 +18,6 @@ ExecStart=/usr/bin/bash -c \
 "if [[ -f /etc/profile.d/etcdctl.sh ]]; then source /etc/profile.d/etcdctl.sh;fi && sudo /usr/bin/docker run --name dd-agent-mesos -h `hostname` \
 -e API_KEY=`etcdctl get /datadog/config/api-key` \
 -e MESOS_HOST=`etcdctl get /environment/CONTROL_ELB` \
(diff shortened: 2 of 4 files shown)


f4tq:
Hold off on merging.  My upstream merge made something wonky.  
Thanks

f4tq:
General system health has been very unstable since merging with upstream.  Suspect combo of too small a control instance size - `m3.medium` -  together with heavy duty processes.
- system-coredump  - mesos-master/meso-slave and java continually dump core crushing m3.medium hosts.  Going bigger today.  Changing `/etc/systemd/coredump.conf`

```
[Coredump]
Storage=none
```

`coredumpctl list` has 50+ cores per control host.  
- cfn-signal jobs - signal completion but then continually boot-roll.  Shouldn't cfn-signal decommission itself i.e `type=OneShot` after fulfilling it's purpose?
- After  merging with upstream and using ETCDCTL_ROOT_PASSWORD, ETCDCTL_READ_PASSWORD, & ETCDCTL_WRITE_PASSWORD, my logs wallpaper with messages like `May 31 15:11:37 ip-10-74-131-154.ec2.internal etcd2[1039]: auth: no authorization provided, checking guest access`
- zookeeper spewing in logs over dropped connections/timeouts which I've seen with Solr when overall system load is too high.  Since so many things rely on zk, I would consider making the zk ensemble separate from the control processing so chain reaction events I'm seeing don't happen.
- All this leads to system-journald running ranking as the top cpu/io process on control nodes.

All and all, I strongly suspect `m3.medium` is too small for control nodes....

anshupande:
`like May 31 15:11:37 ip-10-74-131-154.ec2.internal etcd2[1039]: auth: no authorization provided, checking guest access` this log message is coming after we enabled etcd authentication

PR #228, opened 2016-05-27
added control-proxy service

iderdik:
- Fleet uit for control-proxy
- default container specified

### DIFF
--- a/v3/fleet/control-proxy.service
+++ b/v3/fleet/control-proxy.service
@@ -5,7 +5,7 @@ After=docker.service bootstrap.service zk-health.service
 
 [Service]
 Environment="IMAGE=etcdctl get /images/control-proxy"
-Environment="INTERNAL_CONTROL_ELB=etcdctl get /images/control-proxy"
+Environment="INTERNAL_CONTROL_ELB=etcdctl get /environment/CONTROL_ELB"
 EnvironmentFile=/etc/environment
 User=core
 Type=simple
--- a/v3/fleet/control-proxy.service
+++ b/v3/fleet/control-proxy.service
@@ -1,0 +1,30 @@
+[Unit]
+Description=run apigateway as control proxy
+Requires=docker.service
+After=docker.service bootstrap.service zk-health.service
+
+[Service]
+Environment="IMAGE=etcdctl get /images/control-proxy"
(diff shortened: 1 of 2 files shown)


anshupande:
we can pass this as variable: `etcdctl get /environment/CONTROL_ELB` in environment

v3/fleet/control-proxy.service
```diff
@@ -0,0 +1,30 @@
...
+Restart=always
+RestartSec=4
+TimeoutStartSec=0
+ExecStartPre=/usr/bin/sh -c "docker pull $($IMAGE)"
+ExecStartPre=-/usr/bin/docker kill control-proxy
+ExecStartPre=-/usr/bin/docker rm control-proxy
+ExecStart=/usr/bin/bash -c \
+  '/usr/bin/docker run \
+  --name control-proxy \
+  -e MESOS_MASTER_HOST=http://`etcdctl get /environment/CONTROL_ELB`:5050 \
```
anshupande: we can pass this as variable: `etcdctl get /environment/CONTROL_ELB` in environment

v3/fleet/control-proxy.service
```diff
@@ -0,0 +1,31 @@
+[Unit]
+Description=run apigateway as control proxy
+Requires=docker.service
+After=docker.service bootstrap.service zk-health.service
+
+[Service]
+Environment="IMAGE=etcdctl get /images/control-proxy"
+Environment="INTERNAL_CONTROL_ELB=etcdctl get /images/control-proxy"
```
iderdik: @chr0n1x @matthewdfuller: This value is in the `/etc/environment` file as CONTROL_ELB.  Can I omit this and us `$($CONTROL_ELB)` in the unit below?

matthewdfuller: @iderdik yes, you can reference that, as long as you set your `EnvironmentFile` above

anshupande:
👍

v3/fleet/control-proxy.service
iderdik: It did not work with the environment file set. $($CONTROL_ELB) was not set  - I assume the issue is that the docker run is being shelled out so the env context is different? I dunno. It works as is.

PR #230, opened 2016-06-02
docker initiated logrotation

f4tq:
Docker logs need to be rotated by the container instances themselves when using the default json logger because the container maintains an open file handle to the backing log file. From Coreos, this file can be found in /var/lib/docker/containers/_/_.log. To remove them from Cores though causes overlay and devicemapper corruption (unable to remove layers, images and or instances).
In summary,

All docker run commands have been provisioned with max-file & max-size. The values are pulled from etcd.
A default etcd value is set in init if the user doesn't set one via secrets.json
The external logrotate unit is removed.
The docker cleanup script is changed to use the docker client to remove dead containers and untagged layers i.e. those that show as '' when docker images is run.
Setting secrets.json type=etcd and key of /docker/config/logs-max-size and /docker/config/logs-max-file override the default of max-size=10m and max-file=10 i.e. no more than 10 rotated logs and of no more than 10MB in size.
Finally, the related ticket is adobe-community/issues#9054

Testing with ansible after convergence:
- Setup the [inventory](https://git.corp.adobe.com/fortescu/Mesos4Dexi/tree/master/inventory-ethos-f4tq)
- Checking the journals to make sure all `docker run` commands have targeted options

```
ansible coreos_control:coreos_workers -m raw -a 'bash -c "journalctl --no-pager | grep \"docker run\" | grep  -v max-size"' -i ~/Mesos4Dexi/inventory-ethos-f4tq/ -s -f 10
```
- Make sure systemctl ok

```
ansible coreos_control:coreos_workers:coreos_proxy -i ~/Mesos4Dexi/inventory-ethos-f4tq/  -m raw -a 'bash -c "systemctl list-units | grep -v active"'
```

which shouldn't list anything.  

> Note: remove `-v` to see all the appropriately decorated docker commands
- Test marathon access from each discrete node:

```
ansible coreos_control -m raw -a 'curl -L -v -u admin:password "http://{{ ec2_private_ip_address }}:8080"' -s -i ~/Mesos4Dexi/inventory-ethos-f4tq/ 
```

> Note: change the port/username/password for other well known processes 
- Check that all images are pulled on workers

```
ansible coreos_workers -m raw -a "docker images " -i ~/Mesos4Dexi/inventory-ethos-f4tq/ -s -f 10 -vvv 
```
- Check that all images are running on workers. 

```
ansible coreos_workers -m raw -a "docker ps " -i ~/Mesos4Dexi/inventory-ethos-f4tq/ -s -f 10 -vvv 
```
- Inspect last 20 lines of mesos master journal across your control nodes

```
ansible coreos_control -m raw -a 'bash -c "systemctl --no-pager list-units|grep master|grep service|awk \"{print \\\$1}\"| xargs -n 1 -IXX journalctl --no-pager -u XX | tail -20 "' -i ~/Mesos4Dexi/inventory-ethos-f4tq/ -s -f 10
```
- Inspect last 20 lines of mesos workers journal across your worker nodes

```
ansible coreos_workers -m raw -a 'bash -c "systemctl --no-pager list-units|grep slave|grep service|awk \"{print \\\$1}\"| xargs -n 1 -IXX journalctl --no-pager -u XX | tail -20 "' -i ~/Mesos4Dexi/inventory-ethos-f4tq/ -s -f 10
```

Anyhow, all useful for testing this PR.


ts3ng:
👍  omg this would be so nice to have fixed as i'm tired of replacing my worker nodes.

v3/fleet-local/control/cfn-signal@.service
```diff
@@ -16,7 +16,10 @@ ExecStartPre=/usr/bin/systemctl is-active zk-health.service
 ExecStartPre=/usr/bin/systemctl is-active mesos-master@*
 ExecStartPre=/usr/bin/systemctl is-active marathon@*
 ExecStartPre=/usr/bin/sh -c "docker pull $($IMAGE)"
-ExecStartPre=/usr/bin/sh -c "/usr/bin/docker run --rm\
+ExecStartPre=/usr/bin/sh -c "/usr/bin/docker run \
+--log-opt max-size=10m \
```
bossjones, line 20: should this be `--log-opt max-size=`etcdctl get /docker/config/logs-max-size` ?

v3/fleet-local/control/cfn-signal@.service
```diff
@@ -16,7 +16,10 @@ ExecStartPre=/usr/bin/systemctl is-active zk-health.service
...
+--log-opt max-file=10 \
```
bossjones, line 21: should this be `--log-opt max-file=`etcdctl get /docker/config/logs-max-file` ?

f4tq:
There were 2 cases I didn't make the etcdctl call:
- When bootstrapping etcd
- cfn-signal.
  I reversed out cfn-signal due to intermittent etcd wonkiness I'd experience which would cause the signal to fail.  Enough failures from one node caused the whole stack to tear down.  Since this is short-lived, I hardcoded a value.

chr0n1x:
One thing to note is that @matthewdfuller may have a preference as to where some of this logic should go in ethos-systemd and we should talk to him about it some time next week.

init
```diff
@@ -4,6 +4,8 @@
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 HOMEDIR=$(eval echo "~`whoami`")
 source /etc/environment || :
+# Turn off coredumps that seem to happen on smaller, overloaded instances leading to closed-loop feedback
+sudo /usr/bin/sed -i 's/#Storage=external/Storage=none/' /etc/systemd/coredump.conf
```
chr0n1x, line 8: I kind of want to say that we may want to move this logic into the CFN template userdata. But I don't know.

init
```diff
@@ -20,6 +22,23 @@ if [ "$NODE_ROLE" = "it-hybrid" ]; then
   sudo ${SCRIPTDIR}/itv1/util/etcd2-setup.sh $SCRIPTDIR
 fi
 
+if [ "$NODE_ROLE" = "control" ];then
```
chr0n1x, line 25: This is definitely going to have to change. @matthewdfuller has a post-etcd-setup phase I think in ethos-systemd where this logic can be run as a separate script...?

init
```diff
@@ -30,15 +49,10 @@ fi
...
-# SETUP profile.d
-sudo mkdir /etc/profile.d || :
-sudo cp ${SCRIPTDIR}/$VERSION/profile.d/* /etc/profile.d/. 2>/dev/null || :
-sudo cp --dereference ${SCRIPTDIR}/$VERSION/profile.d/${NODE_ROLE}/* /etc/profile.d/.
-
 # SETUP SCRIPTS
 for script in $(ls ${SCRIPTDIR}/$VERSION/setup|grep -e '.sh$')
 do
-    sudo ${SCRIPTDIR}/$VERSION/setup/${script}
+    sudo -E ${SCRIPTDIR}/$VERSION/setup/${script}
```
chr0n1x, line 55: what's `-E` do?

chr0n1x:
Also, is there any way where we can set the `max-file` & `max-size` in the sh env (in `DOCKER_OPTS` or something)? So that the daemon/client automatically use those values without us having to essentially hard-code values into the units/scripts?

One con of doing this way is the implication that we'll probably have to get those values into the cluster as a part of the CFN spinup so any changes to those values would require an A/B. However, that's not...so bad I guess, since an etcd change for those values would require us to re-spin all of our containers anyway and we might be better off just doing an A/B in either case.

That brings up the question of how our application containers would react to such a constraint on the environment though. :/

f4tq:
@chr0n1x 
- On `sudo -E`:  There must be a mis-conception on how sudo handles the environment.  There are a few units that say 'User=core', the ExecStart="source /etc/profile.d/etcdctl.sh && sudo xxx".  The environment is not passed to sudo without `-E`. 
  To see this in action, visit one of your non-control nodes and run:

```
core@ip-10-74-131-134 sandbox-f4tq-us-east-1c-proxy ~ $ source /etc/profile.d/etcdctl.sh && sudo etcdctl ls
Error:  client: etcd cluster is unavailable or misconfigured
error #0: dial tcp 127.0.0.1:4001: getsockopt: connection refused
error #1: dial tcp 127.0.0.1:2379: getsockopt: connection refused

core@ip-10-74-131-134 sandbox-f4tq-us-east-1c-proxy ~ $ source /etc/profile.d/etcdctl.sh && sudo -E etcdctl ls
/bootstrap.service
/datadog
/splunk
/capcom
/coreos.com
/docker
/etcdctl
/flight-director
/images
/marathon
/sumologic
/zookeeper
/ECR
/klam-ssh
/environment
/mesos
```

Notice that without the `-E`, the environment was not passed for `etcdctl  ls` to work the 1st time but works with `-E` the 2nd time. 
Since I added calls to `etcdctl` in places that didn't have previously have them AND the author's intention clearly is for the environment to be passed AND because I didn't want weird failures, I fixed them in this PR.  
This seemed to work before because a lot of scripts ensure a sane environment for themselves - rightfully.

chr0n1x:
@f4tq good to know, thanks!

f4tq:
@chr0n1x:  I started working on a docker PR on Friday to address providing defaults to the docker daemon for use by clients that don't override defaults.
See [docker PR](https://github.com/adobe-community/issues/issues/9506)

PR #231, opened 2016-06-02
Mount container's root filesystem as read only

anshupande:
The container's root file system should be treated as a 'golden image' and any writes to the rootfilesystem should be avoided. You should explicitly define a container volume for writing. Add a `--read-only` flag to allow the container's root filesystem to bemounted as read only. This can be used in combination with volumes to force a container's process to only write to locations that will be persisted. refer: https://benchmarks.cisecurity.org/tools2/docker/CIS_Docker_1.11.0_Benchmark_v1.0.0.pdf


anshupande:
refer: https://github.com/adobe-community/issues/issues/9285

matthewdfuller:
👍

PR #234, opened 2016-06-08
Zookeeper Bugfix

matthewdfuller:
- Marathon and Chronos require the Zookeeper endpoint to be localhost and not the ELB. This is due to the fact that ELB IPs might change, but Marathon does not re-resolve the new IP using the DNS.


iderdik:
👍

chr0n1x:
👍

PR #236, opened 2016-06-21
hack for IT :(

iderdik:
Greetings and thanks for the PR.  We have a few rules so that everyone enjoys your Pull Request.
## Changelog

Remove units harmful for docker 1.10 on IT only.
## Notes

You must explain how any `docker run` executions your PR introduces handles log accumulation. 

Any docker containers introduced through your commit must not accumulate logs in an unsustainable manner.  In other words, no unmanaged logs that aren't/can't be rotated.  

If your PR relies on the default docker log driver, `json`, then this simply means ensuring all `docker run` commands are decorated with either/both of `--log-opt max-size=XX` / `--log-opt max-file=YY` (see the docker documentation for valid XX/YY values)

If you did, have you ensured that all your `docker run` using the default logger have either `max-size` and/or `max-file` set?

Are there any dependencies on github.com/adobe-platform/infrastructure? NO

If so:

Is an update to the base stack required(rare)? NO

Is an A/B required? NO

Are you dependent on new secrets, or infrastructure in any way? NO

### DIFF
--- a/init
+++ b/init
@@ -122,6 +122,6 @@ if [ "$NODE_ROLE" = "it-hybrid" ]; then
   fleetctl destroy docker-cleanup.service
   fleetctl destroy logrotate.service
   fleetctl destroy update-os.service
-  fleetctl fd-hud.service
+  fleetctl destroy fd-hud.service
 fi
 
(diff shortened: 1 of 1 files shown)


matthewdfuller:
👍

PR #237, opened 2016-07-21
switch to behance memcached container

iderdik:
Greetings and thanks for the PR.  We have a few rules so that everyone enjoys your Pull Request.
## Notes

You must explain how any `docker run` executions your PR introduces handles log accumulation. 

Any docker containers introduced through your commit must not accumulate logs in an unsustainable manner.  In other words, no unmanaged logs that aren't/can't be rotated.  

If your PR relies on the default docker log driver, `json`, then this simply means ensuring all `docker run` commands are decorated with either/both of `--log-opt max-size=XX` / `--log-opt max-file=YY` (see the docker documentation for valid XX/YY values)

If you did, have you ensured that all your `docker run` using the default logger have either `max-size` and/or `max-file` set?

Are there any dependencies on github.com/adobe-platform/infrastructure?

If so:

Is an update to the base stack required(rare)?

Is an A/B required?

Are you dependent on new secrets, or infrastructure in any way?

### DIFF
--- a/v2/fleet-local/it-hybrid/memcached.service
+++ b/v2/fleet-local/it-hybrid/memcached.service
@@ -10,7 +10,7 @@ Restart=always
 RestartSec=5
 TimeoutStartSec=0
 
-ExecStartPre=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && docker pull borja/docker-memcached"
+ExecStartPre=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && docker pull behance/docker-memcached"
 ExecStartPre=-/usr/bin/docker kill memcached
 ExecStartPre=-/usr/bin/docker rm memcached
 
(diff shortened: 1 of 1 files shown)


v2/fleet-local/it-hybrid/memcached.service
```diff
@@ -19,10 +19,10 @@ ExecStart=/usr/bin/sh -c "source /etc/profile.d/etcdctl.sh && \
   echo $($CMD) | xargs docker run \
     --name memcached \
     -p 11211:11211 \
-    borja/docker-memcached"
+    behance/docker-memcached"
 
 ExecStop=-/usr/bin/docker stop memcached
-
+orja
```
chr0n1x: wat

nBerg:
:+1:

chr0n1x:
👍

chr0n1x:
so uh...about squashing commits :P

PR #239, opened 2017-04-06
Remove Sysdig, ECR, Chronos

ko-be:
## Changelog

Remove Sysdig, ECR, Chronos

## Notes

No

Are there any dependencies on github.com/adobe-platform/infrastructure?
No

Is an update to the base stack required(rare)?
No

Is an A/B required?
No

Are you dependent on new secrets, or infrastructure in any way?
No


mannytoledo:
👍

v2/setup/images-control.sh
```diff
@@ -12,7 +12,6 @@ if [ "$(etcdctl get images-control-bootstrapped)" == "true" ]; then
 fi
 etcdctl set images-control-bootstrapped true
 
-etcdctl set /images/chronos      "mesosphere/chronos:chronos-2.4.0-0.1.20150828104228.ubuntu1404-mesos-0.27.0-0.2.190.ubuntu1404"
 etcdctl set /images/fd           "behance/flight-director:latest"
```
bossjones: maybe just in case we can get the exact sha of flight-director just in case "latest" for some reason has moved where it currently is? I know if for some reason flight-director version increases, it'll break on portfolio.

digitalyuki: Good call @bossjones we should fix it to the last supported version of flight-director on mesos-systemd... I remember there was something in mesos-systemd that could not allow a newer version of flight-director.

ko-be: @bossjones @digitalyuki locked flight-director & capcom images.

bossjones:
👍

v2/setup/images-proxy.sh
```diff
@@ -14,7 +14,7 @@ if [ "$(etcdctl get images-proxy-bootstrapped)" == "true" ]; then
 fi
 etcdctl set images-proxy-bootstrapped true
 
-etcdctl set /images/capcom       "behance/capcom:latest"
+etcdctl set /images/capcom       "behance/capcom:8fa735d310d0a46fc3acc34245c8bf63e8447ef4"
 etcdctl set /images/capcom2      "behance/capcom:latest"
```
kangman: @ko-be what is capcom2? should that also have the specific SHA?

digitalyuki: I was going to comment on this as well @ko-be ... I know that `images/capcom2` is set to be `behance/capcom:latest` on the currently running cluster, but we should confirm that's *what it should be set to*, or if we should lock the two containers to the same sha, ask maybe Iz....

ko-be: it is unused as far as i can tell @kangman

v2/setup/images-proxy.sh
```diff
@@ -14,7 +14,7 @@ if [ "$(etcdctl get images-proxy-bootstrapped)" == "true" ]; then
 fi
 etcdctl set images-proxy-bootstrapped true
 
-etcdctl set /images/capcom       "behance/capcom:latest"
+etcdctl set /images/capcom       "behance/capcom:8fa735d310d0a46fc3acc34245c8bf63e8447ef4"
```
mannytoledo: Did you confirm this is the right sha?

mannytoledo:
👍  other than my one comment

v2/setup/images-proxy.sh
ko-be: This sha is running on both stage & prod currently @mannytoledo

kangman:
other than capcom2  👍

digitalyuki:
👍

